PR #13, opened 2022-05-10
feat(csprng): add API to check availability of seeders on current machine

IceTDrinker:
### Resolves:

closes https://github.com/zama-ai/concrete-core-internal/issues/136
closes https://github.com/zama-ai/concrete-core-internal/issues/143

### Description

if a seeder requires specific hardware features or OS features (among
other things) the availability of such a seeder on the current machine can
be checked through a call to `is_available` that was added to the Seeder
trait and implemented for existing seeders

### Checklist 

### DIFF
--- a/concrete-csprng/build.rs
+++ b/concrete-csprng/build.rs
@@ -1,0 +1,106 @@
+// To have clear error messages during compilation about why some piece of code may not be available
+// we decided to check the features compatibility with the target configuration in this script.
+
+use std::collections::HashMap;
+use std::env;
+
+// See https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch for various
+// compilation configuration
+
+// Can be easily extended if needed
+pub struct FeatureRequirement {
+    pub feature_name: &'static str,
+    // target_arch requirement
+    pub feature_req_target_arch: Option<&'static str>,
+    // target_family requirement
+    pub feature_req_target_family: Option<&'static str>,
+}
+
+// We implement a version of default that is const which is not possible through the Default trait
+impl FeatureRequirement {
+    // As we cannot use cfg!(feature = "feature_name") with something else than a literal, we need
+    // a reference to the HashMap we populate with the enabled features
+    fn is_activated(&self, build_activated_features: &HashMap<&'static str, bool>) -> bool {
+        *build_activated_features.get(self.feature_name).unwrap()
+    }
+
+    // panics if the requirements are not met
+    fn check_requirements(&self) {
+        let target_arch = get_target_arch_cfg();
+        if let Some(feature_req_target_arch) = self.feature_req_target_arch {
+            if feature_req_target_arch != target_arch {
+                panic!(
+                    "Feature `{}` requires target_arch `{}`, current cfg: `{}`",
+                    self.feature_name, feature_req_target_arch, target_arch
+                )
+            }
+        }
+
+        let target_family = get_target_family_cfg();
+        if let Some(feature_req_target_family) = self.feature_req_target_family {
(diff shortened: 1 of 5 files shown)


IceTDrinker:
I don't know if updating the release here is relevant (as it's part of the big rework for now)

Not sure about how this should be tested, proposals welcome

IceTDrinker:
Added a build.rs file as discussed, let me know if it looks good to you

concrete-csprng/build.rs
```diff
@@ -0,0 +1,112 @@
...
+
+// const vecs are not yet a thing so use a fixed size array (update the array size when adding
+// requirements)
+static FEATURE_REQUIREMENTS: [FeatureRequirement; 3] = [
+    FeatureRequirement {
+        feature_name: "seeder_x86_64_rdseed",
+        feature_req_target_arch: Some("x86_64"),
+        ..FeatureRequirement::default()
+    },
+    FeatureRequirement {
```
aPere3, line 39: I never saw this use of `..`, but I think I get it: It complements the fields with the ones from `default` right ? To be honest, I think I would have directly written the values here, because it would have made the intent clearer !

IceTDrinker: hmm you are right, I was just thinking of the case where we may have a lot of different requirements and may not have wanted to write None a lot, I can change it if you want 🙂

IceTDrinker: or I can rename the FeatureRequirement::default to FeatureRequirement::no_requirements

concrete-csprng/build.rs
```diff
@@ -0,0 +1,112 @@
...
+        if let Some(feature_req_target_family) = *feature_req_target_family {
+            if feature_req_target_family != target_family {
+                panic!(
+                    "Feature `{}` requires target_family `{}`, current cfg: `{}`",
+                    feature_name, feature_req_target_family, target_family
+                )
+            }
+        }
+    }
+}
```
aPere3, line 112: Here, I think I would have tried to implement that slightly differently.

1. Implement a `is_activated` method on `FeatureRequirement`, which checks if the feature was activated or not in the build.
2. Implement a `check_requirements` method on `FeatureRequirement`, which verifies if the requirements are met by the target. 
3. Rewrite this block as `FEATURE_REQUIREMENTS.iter().filter(FeatureRequirement::is_activated).for_each(FeatureRequirement::check_requirements);`

This would make the intent a bit clearer I think !

aPere3 (approved):
Hello @IceTDrinker,

Thanks for this PR ! It looks good for me ! 

I just made a few minor comments that would, in my opinion, make the code clearer in the build.rs, but nothing too important !

Best,

Alex

concrete-csprng/build.rs
IceTDrinker: Yep, this is me coming from less functional languages I guess 🙂 I will try to rewrite that with this in mind

concrete-csprng/build.rs
aPere3: Well, for now I would prefer to remove `default` altogether to keep things simple :slightly_smiling_face: ! In any case, if the requirements become more complicated in the future, it will be easy to rework that with something in the spirit of `default `!

IceTDrinker: removed the default thing

concrete-csprng/build.rs
IceTDrinker: updated as you requested, the only thing I don't quite like is I need to pass a ref to the (unfortunately) non static HashMap for is_activated, but I think it's generally ok !

agnesLeroy (approved):

PR #17, opened 2022-05-12
Add a parallel engine to the default one

agnesLeroy:
### Description
This PR adds an engine in the default backend that implements the parallel counterpart of the bootstrap key creation.

### Checklist 

### DIFF
--- a/concrete-core/src/backends/default/implementation/engines/lwe_bootstrap_key_creation.rs
+++ b/concrete-core/src/backends/default/implementation/engines/lwe_bootstrap_key_creation.rs
@@ -1,12 +1,15 @@
 use concrete_commons::dispersion::Variance;
 use concrete_commons::parameters::{DecompositionBaseLog, DecompositionLevelCount};
 
+#[cfg(feature = "parallel")]
+use crate::backends::default::implementation::engines::parallel::DefaultParallelEngine;
 use crate::backends::default::implementation::engines::DefaultEngine;
 use crate::backends::default::implementation::entities::{
     GlweSecretKey32, GlweSecretKey64, LweBootstrapKey32, LweBootstrapKey64, LweSecretKey32,
     LweSecretKey64,
 };
 use crate::commons::crypto::bootstrap::StandardBootstrapKey as ImplStandardBootstrapKey;
+use crate::prelude::{GlweSecretKeyEntity, LweSecretKeyEntity};
 use crate::specification::engines::{LweBootstrapKeyCreationEngine, LweBootstrapKeyCreationError};
 
 /// # Description:
@@ -87,11 +90,11 @@ impl LweBootstrapKeyCreationEngine<LweSecretKey32, GlweSecretKey32, LweBootstrap
     ) -> LweBootstrapKey32 {
         let mut key = ImplStandardBootstrapKey::allocate(
             0,
-            output_key.0.key_size().to_glwe_size(),
-            output_key.0.polynomial_size(),
+            output_key.glwe_dimension().to_glwe_size(),
+            output_key.polynomial_size(),
             decomposition_level_count,
             decomposition_base_log,
-            input_key.0.key_size(),
+            input_key.lwe_dimension(),
         );
         key.fill_with_new_key(
             &input_key.0,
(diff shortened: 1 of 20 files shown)


concrete-core/src/backends/default/implementation/engines/lwe_bootstrap_key_creation.rs
```diff
@@ -196,3 +197,196 @@ impl LweBootstrapKeyCreationEngine<LweSecretKey64, GlweSecretKey64, LweBootstrap
...
+        input_key: &LweSecretKey32,
+        output_key: &GlweSecretKey32,
+        decomposition_base_log: DecompositionBaseLog,
+        decomposition_level_count: DecompositionLevelCount,
+        noise: Variance,
+    ) -> LweBootstrapKey32 {
+        let mut key = ImplStandardBootstrapKey::allocate(
+            0,
+            output_key.0.key_size().to_glwe_size(),
+            output_key.0.polynomial_size(),
```
IceTDrinker, line 283: do we need to access the underlying object through .0 to get the glwe_size and polynomial_size ?

concrete-core/src/backends/default/implementation/engines/lwe_bootstrap_key_creation.rs
```diff
@@ -196,3 +197,196 @@ impl LweBootstrapKeyCreationEngine<LweSecretKey64, GlweSecretKey64, LweBootstrap
...
+            decomposition_level_count,
+            decomposition_base_log,
+            input_key.0.key_size(),
```
IceTDrinker, line 286: same question here

concrete-core/src/backends/default/implementation/engines/lwe_bootstrap_key_creation.rs
```diff
@@ -196,3 +197,196 @@ impl LweBootstrapKeyCreationEngine<LweSecretKey64, GlweSecretKey64, LweBootstrap
...
+        decomposition_level_count: DecompositionLevelCount,
+        noise: Variance,
+    ) -> LweBootstrapKey64 {
+        let mut key = ImplStandardBootstrapKey::allocate(
+            0,
+            output_key.0.key_size().to_glwe_size(),
+            output_key.0.polynomial_size(),
+            decomposition_level_count,
+            decomposition_base_log,
+            input_key.0.key_size(),
```
IceTDrinker, line 382: same question for accessing through .0 to the underlying implementation

IceTDrinker (approved):
Hello @agnesLeroy!

Thanks for the PR, I'm not familiar with the private API about filling keys so I cannot comment much on how it is used.

Just a very minor comment about accessing underlying implementations through .0 of some keys to get dimension/sizes which I think can be recovered through function calls.

It really is nitpicking and is not great when it comes to consistency but is probably better to try to hide some implementation details.

So up to you for the approach you prefer 🙂 (consistency vs. being a bit more agnostic of the implementation)

Cheers

concrete-core/src/backends/default/implementation/engines/mod.rs
```diff
@@ -53,6 +53,45 @@ impl AbstractEngine for DefaultEngine {
...
+/// # Note:
+///
+/// There is currently no such case, as the default parallel implementation is not expected to
+/// undergo major issues unrelated to FHE.
+#[derive(Debug)]
+pub enum DefaultParallelError {}
+
+impl Display for DefaultParallelError {
+    fn fmt(&self, _f: &mut Formatter<'_>) -> std::fmt::Result {
+        match *self {}
```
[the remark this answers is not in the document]
agnesLeroy: I think it's just necessary because the enum is empty (see https://www.reddit.com/r/rust/comments/43j5fk/what_is_the_meaning_of_match_self/ for example)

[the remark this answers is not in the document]
IceTDrinker: ok thanks!

concrete-core/src/backends/default/implementation/engines/lwe_bootstrap_key_creation.rs
agnesLeroy: No it's not necessary you're right, we should try to hide the implementation details as much as we can.

agnesLeroy:
@IceTDrinker I had to make some changes to include the parallel engine in a module that's only activated with the parallel feature. Maybe there's a nicer way of doing it though :thinking:

IceTDrinker:
> @IceTDrinker I had to make some changes to include the parallel engine in a module that's only activated with the parallel feature. Maybe there's a nicer way of doing it though 🤔

Hmm, there is always the possibility of creating a new folder for the default_engine_parallel, but then our project structure would be even more compilcated 🤔

IceTDrinker:
added a correction in a doc string where there was a `DefaultEngine` that was not converted to a `DefaultParallelEngine`

PR #6, opened 2022-04-28
Feature/seeded ciphertexts

agnesLeroy:
### Resolves: 
- zama-ai/concrete-core-internal#80

### Checklist 

(Use '[x]' to check the checkboxes, or submit the PR and then click the checkboxes)

* [x] Tests for the changes have been added (for bug fixes / features)
* [x] Docs have been added / updated (for bug fixes / features)
* [x] The PR description links to the related issue (to link an issue, use '#XXX'.)
* [x] The tests on AWS have been launched and are successful (apply the `aws_test` to the PR to launch the tests on AWS)
* [ ] The draft release description has been updated
* [x] Check for breaking changes and add them to commit message following the conventional commit [specification][conventional-breaking]

### Requires: 
- zama-ai/concrete/pull/169

[conventional-breaking]: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer

PR #24, opened 2022-05-23
Feature/seeded lwe ciphertext

IceTDrinker:
### Resolves:

refs https://github.com/zama-ai/concrete-core-internal/issues/80
closes https://github.com/zama-ai/concrete-core-internal/issues/166

### Description

Rework with current code base of: https://github.com/zama-ai/concrete-core/pull/6 for the LweSeededCiphertext.

This PR includes a bit more than future PRs:
- some additional entry points/functionality for the CSPRNG (efficient shift mostly)
- the same entry points exposed in the commons from concrete-core
- the actual implementation in commons (ex-private)
- specs and implementations for LweSeededEntity and engines
- fixtures for new engines

I would recommend reviewing commit by commit as I tried to keep commits focused on a single theme

The main reason for the shift to exist is to be able to do efficient decompression of a LweSeededCiphertext in the future LweSeededList, basically all CTs will share a seed in a list but have different shifts to re-use the same ByteRandomGenerator over and over again (avoids having to instantiate a new one each time) when decompressing the whole list. The shift is byte based, this makes the shift easier to use in a context where one would re-use the same generator for a lot of seeded types.

I will update the release drafts once we have ironed out the first kinks in the PR (naming/which functions are needed e.g.)

### Checklist 


concrete-csprng/src/generators/mod.rs
```diff
@@ -70,18 +87,40 @@ pub trait RandomGenerator: Iterator<Item = u8> {
...
+    /// `min(2¹²⁸-1, generated_bytes)`.
+    fn generated_bytes_so_far(&self) -> ByteCount;
+
     /// Returns the next byte of the stream, if the generator did not yet reach its bound.
     fn next_byte(&mut self) -> Option<u8> {
         self.next()
     }
 
+    /// Skips n bytes without consuming the generator like `Iterator::skip` does.
+    fn shift_n(&mut self, n: usize) -> Result<(), GeneratorError> {
```
agnesLeroy, line 116: Shouldn't we name the fn `skip` then? It seems to me that `shift` conveys the idea you can move the index both ways, whereas we actually only support a "forward shift", right?

concrete-csprng/src/generators/mod.rs
```diff
@@ -70,18 +87,40 @@ pub trait RandomGenerator: Iterator<Item = u8> {
...
+        if n > 0 {
+            // nth(n) skips n + 1
+            self.nth(n - 1).ok_or(GeneratorError::ShiftTooLarge(n))?;
+        }
```
agnesLeroy, line 120: Should we return an error when n is not > 0?

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
```diff
@@ -0,0 +1,193 @@
...
+
+/// A seeded ciphertext encrypted using the LWE scheme.
+#[cfg_attr(feature = "serde_serialize", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LweSeededCiphertext<Scalar> {
+    pub(crate) body: LweBody<Scalar>,
+    pub(crate) lwe_dimension: LweDimension,
+    // Use a proxy to serialize/deserialize Seed to avoid adding a serde dependency to
+    // concrete-csprng
+    #[cfg_attr(feature = "serde_serialize", serde(with = "SeedSerdeDef"))]
```
agnesLeroy, line 24: Maybe we could want to introduce serde as a dependency in concrete-csprng? We don't have users other than core for that crate atm, so it's hard to say :sweat_smile:

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub(crate) seed: Seed,
+    // shift is the sequence number of the first byte to generate on a freshly seeded CSPRNG to
+    // properly generate the mask of the ciphertext. technically this should be a u128, but usize
+    // is plenty to generate bytes, also `Iterator` APIs take usize.
+    pub(crate) shift: usize,
```
agnesLeroy, line 29: Maybe this should be named `generator_index` to be more specific?

concrete-core/src/commons/crypto/secret/generators/encryption.rs
```diff
@@ -34,16 +34,29 @@ impl<G: ByteRandomGenerator> EncryptionRandomGenerator<G> {
...
 
+    /// Returns the number of bytes that have been generated for the mask so far if it fits in a
+    /// usize. On a fresh generator this returns 0.
+    pub fn generated_bytes_so_far(&self) -> Option<usize> {
+        self.mask.generated_bytes_so_far()
+    }
+
+    /// Skip n bytes for the mask generator. Depending on the underlying implementation of the
+    /// CSPRNG this may short circuit avoiding the cost of generating n bytes.
+    pub fn shift_n(&mut self, n: usize) {
```
agnesLeroy, line 56: Maybe this should be named `skip_n`

concrete-core/src/commons/crypto/secret/lwe.rs
```diff
@@ -304,6 +305,34 @@ where
         LweDimension(self.as_tensor().len())
     }
 
+    fn fill_lwe_mask_and_body_for_encryption<OutputCont, Scalar, Gen>(
+        &self,
+        output_body: &mut LweBody<Scalar>,
+        output_masks: &mut LweMask<OutputCont>,
```
agnesLeroy, line 311: I would name is `output_mask` here: the mask refers to the whole vector of values used to generate the body.

concrete-core/src/commons/crypto/secret/lwe.rs
```diff
@@ -304,6 +305,34 @@ where
...
+        encoded: &Plaintext<Scalar>,
+        noise_parameters: impl DispersionParameter,
```
agnesLeroy, line 313: Here can't we just pass a noise variance instead of using the DispersionParameter type?

concrete-core/src/backends/default/implementation/entities/lwe_seeded_ciphertext.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    fn lwe_dimension(&self) -> LweDimension {
+        self.0.lwe_size().to_lwe_dimension()
+    }
+
+    fn seed(&self) -> Seed {
+        self.0.get_seed()
+    }
+
+    fn shift(&self) -> usize {
```
agnesLeroy, line 27: Maybe shift could be renamed into `generator_index` or `generator_byte_index`, something of the kind?

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
```diff
@@ -0,0 +1,46 @@
...
+engine_error! {
+    LweSeededCiphertextEncryptionError for LweSeededCiphertextEncryptionEngine @
+}
+
+/// A trait for engines encrypting seeded LWE ciphertexts.
+///
+/// # Semantics
+///
+/// This [pure](super#operation-semantics) operation generates an LWE ciphertext containing the
+/// encryption of the `input` plaintext under the `key` secret key.
```
agnesLeroy, line 18: Here there should be a `Formal definition` section with the description of the encryption of a seeded ciphertext.

concrete-core/src/specification/entities/lwe_seeded_ciphertext.rs
```diff
@@ -0,0 +1,23 @@
+use crate::commons::math::random::Seed;
+use crate::specification::entities::markers::{KeyDistributionMarker, LweSeededCiphertextKind};
+use crate::specification::entities::AbstractEntity;
+use concrete_commons::parameters::LweDimension;
+
+/// A trait implemented by types embodying a seeded LWE ciphertext.
+///
+/// A seeded LWE ciphertext is associated with a
+/// [`KeyDistribution`](`LweSeededCiphertextEntity::KeyDistribution`) type, which conveys the
+/// distribution of the secret key it was encrypted with.
```
agnesLeroy, line 10: Here there should be a description of what a seeded LWE ciphertext is.

concrete-core/src/backends/default/implementation/engines/mod.rs
```diff
@@ -33,6 +33,7 @@ impl Error for DefaultError {}
 pub struct DefaultEngine {
     secret_generator: ImplSecretRandomGenerator<SoftwareRandomGenerator>,
     encryption_generator: ImplEncryptionRandomGenerator<SoftwareRandomGenerator>,
+    seeder: Box<dyn Seeder>,
```
agnesLeroy, line 36: Why do we need to add the seeder to the engine? Isn't it enough to have the two random generators?

concrete-core/src/specification/engines/lwe_seeded_to_lwe_ciphertext_transmutation.rs
```diff
@@ -0,0 +1,38 @@
...
+engine_error! {
+    LweSeededToLweCiphertextTransmutationEngineError for LweSeededToLweCiphertextTransmutationEngine @
+}
+
+/// A trait for engines transmuting LWE seeded ciphertexts into LWE ciphertexts.
+///
+/// # Semantics
+///
+/// This [pure](super#operation-semantics) operation moves the existing LWE seeded ciphertext into a
+/// LWE ciphertext.
```
agnesLeroy, line 15: Here there should be a `Formal definition` to describe what the transmutation does.

agnesLeroy (commented):
Hello @IceTDrinker! Thanks a lot for this very neat piece of work! I have some comments on the PR, the main things being:
- I guess it's ok not to introduce a dependency to serde in the csprng, but maybe we'll want to have it at some point? What would be your opinion on that?
- regarding the naming, I suggest to use `skip` instead of `shift` in the generator and `generator_byte_index` or `generator_index` in the seeded ciphertext itself. What do you think? @damienligier what do you think?
- there is some crypto documentation missing: @damienligier how would you like us to proceed? Shoud @IceTDrinker try to write something and you review it? Or is it better that you write something in the obsidian documentation base and send it to him?

concrete-csprng/src/generators/mod.rs
IceTDrinker: So the tricky thing here is that the RandomGenerator implements the Iterator trait which has a function called skip which would make naming this skip ambiguous.

To make things worse, the skip method on an iterator would consume the object making accessing the original generator impossible (at least I could not find a way to make it work the way I wanted).

Happy to discuss naming options, but skip because of Iterator is probably not a good idea 🙂

concrete-csprng/src/generators/mod.rs
IceTDrinker: Hmm that's a good question, I took inspiration from the skip method of the iterator trait and they had something that looked like this

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
IceTDrinker: I preferred to be careful on that, it's easy to change if needed 🙂 

We can feature gate it with a serde_serialize, I just did not want to put a constraint on csprng because of concerns not linked to csprng.

If we think it makes sense to add the ability to serialize in csprng, it's fine by me !

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
IceTDrinker: open to naming, I could not think of good naming while writing it and stuck to the original from the PR being reworked.

Strictly it is the "sequence number of the first byte to generate for the mask of the seeded entity"

concrete-core/src/commons/crypto/secret/generators/encryption.rs
IceTDrinker: I can do skip_n 🙂

concrete-core/src/commons/crypto/secret/lwe.rs
IceTDrinker: fine by me, I was re-using the original PR naming 🙂

concrete-core/src/commons/crypto/secret/lwe.rs
IceTDrinker: re-used from original PR, it can be changed easily

concrete-core/src/backends/default/implementation/entities/lwe_seeded_ciphertext.rs
IceTDrinker: generator_byte_index is probably the most accurate one, I'll use it then !

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
IceTDrinker: I did not see one in lwe_ciphertext_encryption.rs that's why I did not add one. As we had a discussion a few weeks back about avoiding empty formal definitions I did not try furter. Let me know what we should do about this (creating issues for missing formal def e.g.).

I can add the definition here

concrete-csprng/src/generators/mod.rs
agnesLeroy: Oh ok, then let's keep `shift`, I don't have a better idea :sweat_smile:

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
agnesLeroy: It was just an open question on my side :slightly_smiling_face: I guess for now we can keep what you did: as you say it'll be easy to change if we have requests to support serialization in the csprng itself.

concrete-core/src/commons/crypto/secret/generators/encryption.rs
agnesLeroy: Let's keep `shift` here to be consistent with the csprng naming.

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
agnesLeroy: Actually I merged the documentation PR yesterday! So when you rebase on main you'll see all the formal definitions & entity documentation. :smile:

IceTDrinker: Ah great 😄 will be in the next set of patches then !

concrete-csprng/src/generators/mod.rs
agnesLeroy: Indeed they don't specifically return an error when n is not > 0. I guess it's ok like this then :sweat_smile:

concrete-core/src/backends/default/implementation/engines/mod.rs
agnesLeroy: Alright then! Maybe we should have a small documentation explaining the purpose of the two generators and the seeder.

IceTDrinker: agreed

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
IceTDrinker: discussion here: https://github.com/zama-ai/concrete-core/pull/24#discussion_r880165829

concrete-core/src/specification/engines/lwe_seeded_to_lwe_ciphertext_transmutation.rs
IceTDrinker: Same as the other Forma definition, I was a bit optimistic at first, there does not seem to be a written example of Formal definition to take inspiration from.

Let's find some time to discuss how this should be presented

concrete-core/src/backends/default/implementation/engines/mod.rs
IceTDrinker: Docstring added, let me know if it's ok !

agnesLeroy: Thanks! maybe for the seeder comment I would explain why we can't just reuse the encryption generator?

IceTDrinker: I was thinking about this again, it’s probably possible to store only a Seed in the DefaultEngine, re-using the encryption generator would require being able to set the state of the generator to whatever state we want. We don’t currently have that in CSPRNG, I don’t know if it could be interesting to consider adding the ability to set the CSPRNG in the state we want

agnesLeroy: Yes that seems like a nice feature to have. Having only the seed stored in the engine would look better in my opinion.

IceTDrinker: Yep agreed, plus now that I re-checked how everything is done nowadays, we have the encryption CSPRNG exactly yo avoid  having to re create some generators (which probably was not the case in the original seeded PR)

concrete-core/src/commons/crypto/secret/lwe.rs
IceTDrinker: refactor issue here: https://github.com/zama-ai/concrete-core-internal/issues/26

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
IceTDrinker: As discussed I will prepare a draft proposal

IceTDrinker:
After discussing with @agnesLeroy we came to the conclusion that it would be easier to complete this PR after the Aesni backend is merged #26 so that we can have a mechanism to provide efficient random generators to the seeded encryption/expansion functions if they are available.

This will probably require something like a factory being available in the engines to provide said efficient generators.

concrete-core/src/backends/default/implementation/engines/mod.rs
IceTDrinker: Potentially not a good idea, this means all seeded entities share the same seed and for an engine to be able to encrypt/expand seeded entities properly it would mean providing a known seed when instantiating it.

This means greater freedom but also more opportunities for errors.

To be discussed, those are design decisions we need to make together

IceTDrinker: Instantiating an engine with a Seed ahead of time means it cannot process seeded entities with other seeds which is not a good idea.

So better to instantiate encryption PRNG when required. See https://github.com/zama-ai/concrete-core/pull/24#issuecomment-1137269741

Encryption PRNG need a seeder when being created to seed the private PRNG used for errors.

This seeder is therefore required.

Closing this conversation 🙂

IceTDrinker:
pushed a proposal for Formal definitions as discussed

IceTDrinker:
updated with latest code base

agnesLeroy (commented):
Hi @IceTDrinker! Just starting to look at the PR again, I'm asking myself one thing: in order to be correctly decompressed, the same csprng has to be used on the client and on the server, right? With the current implementation, unless I 'm missing something, this is not something we can ensure, can we?

IceTDrinker:
> Hi @IceTDrinker! Just starting to look at the PR again, I'm asking myself one thing: in order to be correctly decompressed, the same csprng has to be used on the client and on the server, right? With the current implementation, unless I 'm missing something, this is not something we can ensure, can we?

Actually both implement the same algorithm 🙂 the AES block cipher, so as long as they have the algorithm implementation we are good.

If we start having different families of CSPRNG in concrete-csprng then yes we will have to change how this is done !

agnesLeroy:
> > Hi @IceTDrinker! Just starting to look at the PR again, I'm asking myself one thing: in order to be correctly decompressed, the same csprng has to be used on the client and on the server, right? With the current implementation, unless I 'm missing something, this is not something we can ensure, can we?
> 
> Actually both implement the same algorithm slightly_smiling_face the AES block cipher, so as long as they have the algorithm implementation we are good.
> 
> If we start having different families of CSPRNG in concrete-csprng then yes we will have to change how this is done !

Thank you for the clarification!

concrete-core/src/specification/engines/lwe_seeded_to_lwe_ciphertext_transmutation.rs
agnesLeroy: I think we can resolve these documentation discussions, right?

agnesLeroy (approved):
@IceTDrinker for me the PR looks good, thanks a lot for your work! :smiley:

concrete-core/src/commons/crypto/secret/generators/encryption.rs
```diff
@@ -34,16 +34,29 @@ impl<G: ByteRandomGenerator> EncryptionRandomGenerator<G> {
 
     // Allows to seed the noise generator. For testing purpose only.
     #[allow(dead_code)]
+    #[cfg(test)]
     pub(crate) fn seed_noise_generator(&mut self, seed: Seed) {
```
aPere3, line 38: Can't we remove the `allow(dead_code)` if we do that ? That would probably be a better a way to handle this kind of situation btw !

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub(crate) lwe_dimension: LweDimension,
+    // Use a proxy to serialize/deserialize Seed to avoid adding a serde dependency to
+    // concrete-csprng
+    #[cfg_attr(feature = "serde_serialize", serde(with = "SeedSerdeDef"))]
+    pub(crate) seed: Seed,
+    // generator_byte_index is the sequence number of the first byte to generate on a freshly
+    // seeded CSPRNG to properly generate the mask of the ciphertext. Technically this should
+    // be a u128, but usize is plenty to generate bytes, also `Iterator` APIs take usize.
+    pub(crate) generator_byte_index: usize,
+}
```
aPere3, line 30: Reading this, I was wondering if we shouldn't have a special type `CryptoSeed(u128, usize)`, that packs the csprng seed, and the shift. Because, if I am not mistaken, those always go hand in hand. And really, a ciphertext _seed_ (reading the informations needed to properly generate the ciphertext), is the combination.

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// use concrete_commons::parameters::{LweDimension, LweSize};
+    /// use concrete_core::commons::crypto::lwe::{LweBody, LweSeededCiphertext};
+    /// use concrete_core::commons::math::random::Seed;
+    /// let ciphertext = LweSeededCiphertext::allocate(LweDimension(3), Seed(42), 37);
+    /// assert_eq!(*ciphertext.get_body(), LweBody(0_u8));
+    /// assert_eq!(ciphertext.lwe_size(), LweSize(4));
+    /// assert_eq!(ciphertext.get_seed(), Seed(42));
+    /// assert_eq!(ciphertext.get_generator_byte_index(), 37);
+    /// ```
+    pub fn allocate(lwe_dimension: LweDimension, seed: Seed, shift: usize) -> Self {
```
aPere3, line 47: Whether or not we do this `CryptoSeed` stuff, I would prefer to have a newtype for those shifts. We tried hard to avoid passing raw usizes like that in the rest of the api, to make the things clearer at call site. I think this is something we should keep doing !

concrete-core/src/specification/entities/lwe_seeded_ciphertext.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// Returns the LWE dimension of the ciphertext.
+    fn lwe_dimension(&self) -> LweDimension;
+
+    /// Returns the seed used to generate the mask of the LWE ciphertext during encryption.
+    fn seed(&self) -> Seed;
+
+    /// Returns the shift used to generate the mask of the LWE ciphertext during encryption.
+    fn generator_byte_index(&self) -> usize;
+}
```
aPere3, line 30: Do we really need a separate entity trait for those ? I had the feeling that implementing `LweCiphertextEntity` for the seeded types, and the `LweCiphertextConversionEngine` would be enough, but maybe I am missing something ?

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// For the _general_ safety concerns regarding this operation, refer to the different variants
+    /// of [`LweSeededCiphertextEncryptionError`]. For safety concerns _specific_ to an
+    /// engine, refer to the implementer safety section.
+    unsafe fn encrypt_lwe_seeded_ciphertext_unchecked(
+        &mut self,
+        key: &SecretKey,
+        input: &Plaintext,
+        noise: Variance,
+    ) -> Ciphertext;
+}
```
aPere3, line 67: Again, here, wouldn't it be enough to use the existing encryption trait ?

concrete-core/src/backends/default/implementation/engines/mod.rs
aPere3: I did not understand all the details of the discussion above, but I don't think we require a `seeder` here. We can just create a `ConstantSeeder` at call site which returns the seed given to it to satisfy the api. The seed would be generated using the existing prngs.

IceTDrinker:
fixed some doc strings

concrete-csprng/src/generators/aes_ctr/generic.rs
```diff
@@ -114,6 +114,19 @@ impl<BlockCipher: AesBlockCipher> AesCtrGenerator<BlockCipher> {
...
+    /// Note :
+    /// ------
+    ///
+    /// Note that `ByteCount` uses the `u128` datatype to store the byte count. Unfortunately, the
+    /// number of generated bytes is in ⟦0;2¹³² -1⟧. When the number is greater than 2¹²⁸ - 1,
+    /// we saturate the count at 2¹²⁸ - 1.
+    pub fn generated_bytes_so_far(&self) -> ByteCount {
+        TableIndex::distance(&self.state.table_index(), &TableIndex::FIRST).unwrap()
+    }
+
```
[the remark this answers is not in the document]
aPere3: Also, I don't see where this is used. Could you show me ?

concrete-core/src/specification/engines/lwe_seeded_to_lwe_ciphertext_transmutation.rs
```diff
@@ -0,0 +1,55 @@
...
+    ///
+    /// # Safety
+    /// For the _general_ safety concerns regarding this operation, refer to the different variants
+    /// of [`LweSeededToLweCiphertextTransmutationEngineError`].
+    /// For safety concerns _specific_ to an engine, refer to the implementer safety section.
+    unsafe fn transmute_lwe_seeded_ciphertext_to_lwe_ciphertext_unchecked(
+        &mut self,
+        lwe_seeded_ciphertext: InputCiphertext,
+    ) -> OutputCiphertext;
+}
```
aPere3, line 55: Again, I find adding this to the spec kind of heavy, knowing we have `LweCiphertextConversionEngine` around. 

Also, I think that in the other places, we used the `transmute` word differently. The idea was that transmute was only re-interpreting existing data, and as such it was a free operation. Here, we perform an heavy operation in the background, which is kind of different I think.

aPere3 (changes requested):
Hello @IceTDrinker !

Thanks for your work on this PR ! 

I have a few questions / comments on this one. Hope it is not too late to discuss them ! 

Cheers,

Alex

concrete-csprng/src/generators/aes_ctr/generic.rs
```diff
@@ -186,10 +186,21 @@ impl<BlockCipher: AesBlockCipher> Iterator for AesCtrGenerator<BlockCipher> {
...
+    // advance_by is unstable, the current recommendation is to implement nth to advance an iterator
+    // here we will avoid computing AES a lot of times and just move the state to the proper byte.
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
         if self.state.table_index() >= self.last {
             None
         } else {
-            match self.state.increment() {
+            // nth is zero indexed, so increase by 1
+            match self.state.increase(n + 1) {
                 ShiftAction::OutputByte(BufferPointer(ptr)) => Some(self.buffer[ptr]),
```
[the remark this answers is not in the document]
aPere3: Talking about the `nth` api ...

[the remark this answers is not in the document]
IceTDrinker: it is used in the shift operation concrete-csprng/src/generators/mod.rs line 116

Naming is a bit of a pain around these operations, let me know how the naming looks to you !

concrete-core/src/commons/crypto/secret/generators/encryption.rs
IceTDrinker: yes you are right this allow dead_code becomes redundant !

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
IceTDrinker: I think CryptoSeed may be misleading, perhaps a `MaskSeed` ?

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
IceTDrinker: Agreed ! so should we use a combined struct or just a new type for the shift ?

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
aPere3: Yes why not ! I don't really care, as long as its different from `Seed` so that we are not mistaken :smile:

concrete-core/src/specification/entities/lwe_seeded_ciphertext.rs
IceTDrinker: well you can't do any operation on seeded entities unless you decompress them, to me it warranted a new entity trait as the semantics is really different, I could be wrong 🙂

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
aPere3: For me its best to use a single type as discussed above !

concrete-core/src/backends/default/implementation/engines/mod.rs
IceTDrinker: Sorry I don't understand what you mean @aPere3 🤔

IceTDrinker: The reason we need a seeder is that to encrypt an LWE ciphertext with any seed we require an EncryptionRandomGenerator with a seed for the mask *AND* a seeder for the private noise generator

concrete-csprng/src/generators/aes_ctr/generic.rs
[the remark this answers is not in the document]
IceTDrinker: hmm actually I'm mistaken, it seems I'm not using it as I'm putting 0 every time, maybe it's not needed anymore, I'll try to see what's what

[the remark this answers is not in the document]
IceTDrinker: ok so I now remember (sorry it has been about a month 😅 ) for individual ciphertexts it's not really useful, it's useful for decompressing specific ciphertexts from a vector, to be able to indicate which byte index to use for that ciphertext

concrete-core/src/specification/engines/lwe_seeded_to_lwe_ciphertext_transmutation.rs
IceTDrinker: yes I think so ! I'll do a pass on discussions once I push the reworked PR with the design we discussed !

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
IceTDrinker: resolving, we'll open new conversations if needed

concrete-core/src/specification/entities/lwe_seeded_ciphertext.rs
IceTDrinker: resolving we'll open new discussions if needed

concrete-csprng/src/generators/aes_ctr/generic.rs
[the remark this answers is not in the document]
IceTDrinker: I think you found the nth and this was linked to the nth thing ?

concrete-csprng/src/generators/aes_ctr/generic.rs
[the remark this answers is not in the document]
IceTDrinker: removed in the new iteration

concrete-core/src/commons/crypto/secret/generators/encryption.rs
IceTDrinker: done

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
IceTDrinker: I went for CompressionSeed as discussed, which currently just wraps a Seed, but at least clearer to us, users of the internal API

concrete-core/src/specification/entities/lwe_seeded_ciphertext.rs
IceTDrinker: As discussed in https://github.com/zama-ai/concrete-core-internal/issues/241 we agreed new types were a better fit this time

concrete-core/src/commons/crypto/lwe/seeded_ciphertext.rs
IceTDrinker: Done, removed the shift as it was not really needed in our use cases

Renamed CryptoSeed to CompressionSeed

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
IceTDrinker: as we discussed, better with new traits this time from https://github.com/zama-ai/concrete-core-internal/issues/241

IceTDrinker:
> Pushed an updated version of the PR without the CSPRNG-backed seeder, this felt like something a bit bigger/different, I'll add it in a next version

or maybe a different PR altogether as it modifies the way the DefaultEngine works for seeding @aPere3

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
```diff
@@ -0,0 +1,67 @@
...
+
+/// A trait for engines encrypting seeded LWE ciphertexts.
+///
+/// # Semantics
+///
+/// This [pure](super#operation-semantics) operation generates an LWE ciphertext containing the
+/// encryption of the `input` plaintext under the `key` secret key.
+///
+/// # Formal Definition
```
damienligier, line 20:
```
  LWE Encryption
  inputs:
$\mathsf{pt}\in\mathbb{Z}_q$: a plaintext
$\vec{s}\in\mathbb{Z}_q^n$: a secret key
$\mathsf{seed}$: a public seed
$G$: a CSPRNG working with seeds from $\mathcal{S}$
$\mathcal{D}_{\sigma^2,\mu}$: a normal distribution of variance $\sigma^2$ and a mean $\mu$



  outputs:
$\tilde{\mathsf{ct}} = \left( \mathsf{seed} , b\right) \in \mathsf{SeededLWE}^n_{\vec{s}, G}(
   \mathsf{pt})\subseteq \mathcal{S}\times \mathbb{Z}_q$: a seeded LWE ciphertext



  algorithm:
 1. uniformly sample $n$ integers in $\mathbb{Z}_q$ from $G$ with the seed $\mathsf{seed}\in\mathcal{S}$ and store them in $\vec{a}\in\mathbb{Z}^n_q$
 2. sample an integer error term $e \hookleftarrow \mathcal{D}_{\sigma^2,\mu}$
 3. compute $b = \left\langle \vec{a} , \vec{s} \right\rangle + \mathsf{pt} + e \in\mathbb{Z}_q$
 4. output $\left( \mathsf{seed} , b\right)$
 ```

**Review by damienligier — CHANGES_REQUESTED** [review]
thanks for the pr. i have a few questions. i updated the formal def.

File: concrete-core-fixture/src/generation/prototypes/lwe_seeded_ciphertext.rs [hunk 22]
```diff
@@ -0,0 +1,23 @@
...
+use concrete_core::prelude::{LweSeededCiphertext32, LweSeededCiphertext64};
+
+/// A trait implemented by LWE seeded ciphertext prototypes.
+pub trait LweSeededCiphertextPrototype {
+    type KeyDistribution: KeyDistributionMarker;
+    type Precision: IntegerPrecision;
+}
+
+/// A type representing the prototype of a 32 bit binary LWE seeded ciphertext entity.
+pub struct ProtoBinaryLweSeededCiphertext32(pub(crate) LweSeededCiphertext32);
```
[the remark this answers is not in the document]
IceTDrinker: It is the current design pattern in concrete-core-fixture, not sure why it was done this way 🤔 I'm guessing it's easier to make the distinction once we have more key distributions supported in core

concrete-core-fixture/src/generation/prototypes/lwe_seeded_ciphertext.rs
```diff
@@ -0,0 +1,23 @@
...
+impl LweSeededCiphertextPrototype for ProtoBinaryLweSeededCiphertext32 {
+    type KeyDistribution = BinaryKeyDistribution;
+    type Precision = Precision32;
+}
+
+/// A type representing the prototype of a 64 bit binary LWE seeded ciphertext entity.
+pub struct ProtoBinaryLweSeededCiphertext64(pub(crate) LweSeededCiphertext64);
```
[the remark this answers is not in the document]
IceTDrinker: same as above

concrete-core/src/backends/default/implementation/engines/mod.rs
```diff
@@ -39,8 +39,13 @@ type ActivatedRandomGenerator = AesniRandomGenerator;
 type ActivatedRandomGenerator = SoftwareRandomGenerator;
 
 pub struct DefaultEngine {
+    /// A CSPRNG to generate secret key material.
```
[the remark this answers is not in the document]
IceTDrinker: yes certainly, will update

concrete-core/src/backends/default/implementation/engines/mod.rs
```diff
@@ -39,8 +39,13 @@ type ActivatedRandomGenerator = AesniRandomGenerator;
...
     secret_generator: ImplSecretRandomGenerator<ActivatedRandomGenerator>,
+    /// A CSPRNG to generate secret material for encryption like masks and errors.
```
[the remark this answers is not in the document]
IceTDrinker: The EncryptionRandomGenerator contains two CSPRNG, a mask one that is explicitely seeded by the user (us, core devs) and a noise generator which is seeded using a seeder to make sure the same seed is not re-used by error on the noise generator

concrete-core/src/backends/default/implementation/entities/lwe_seeded_ciphertext.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// A structure representing a seeded LWE ciphertext with 32 bits of precision.
+#[cfg_attr(feature = "serde_serialize", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LweSeededCiphertext32(pub(crate) ImplLweSeededCiphertext<u32>);
+impl AbstractEntity for LweSeededCiphertext32 {
+    type Kind = LweSeededCiphertextKind;
+}
+impl LweSeededCiphertextEntity for LweSeededCiphertext32 {
+    type KeyDistribution = BinaryKeyDistribution;
```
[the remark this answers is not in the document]
IceTDrinker: Today in core the entities of the public API are all associated with the BinaryKeyDistribution and the names don't contain "binary", I followed this pattern here

concrete-core/src/backends/default/implementation/entities/lwe_seeded_ciphertext.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// A structure representing a seeded LWE ciphertext with 64 bits of precision.
+#[cfg_attr(feature = "serde_serialize", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LweSeededCiphertext64(pub(crate) ImplLweSeededCiphertext<u64>);
+impl AbstractEntity for LweSeededCiphertext64 {
+    type Kind = LweSeededCiphertextKind;
+}
+impl LweSeededCiphertextEntity for LweSeededCiphertext64 {
+    type KeyDistribution = BinaryKeyDistribution;
```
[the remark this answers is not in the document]
IceTDrinker: same as above

concrete-core/src/commons/crypto/secret/lwe.rs
```diff
@@ -304,6 +304,34 @@ where
...
+        Self: AsRefTensor<Element = Scalar>,
+        OutputCont: AsMutSlice<Element = Scalar>,
+        Scalar: UnsignedTorus,
+        Gen: ByteRandomGenerator,
+    {
+        // generate a uniformly random mask
+        generator.fill_tensor_with_random_mask(output_mask);
+
+        // generate an error from the normal distribution described by std_dev
+        output_body.0 = generator.random_noise(noise_parameters);
```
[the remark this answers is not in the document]
IceTDrinker: generator here is an EncryptionRandomGenerator the first fill uses the mask generator seeded with a seed passed by the developer, the noise generator is used for random_noise and is seeded with a seeder so that the seed used is not a mistakenly re-used one by the developer

concrete-core/src/commons/crypto/secret/lwe.rs
```diff
@@ -354,21 +382,101 @@ where
...
+            EncryptionRandomGenerator::<Gen>::new(output.get_seed(), seeder.as_mut());
+        // Shift the generator to match the state required by the ciphertext
+        generator.shift(output.get_generator_byte_index());
 
-        // add the encoded message
-        output_body.0 = output_body.0.wrapping_add(encoded.0);
+        let mut output_mask = LweMask::from_container(vec![Scalar::ZERO; self.key_size().0]);
+        let output_body = output.get_mut_body();
+
+        self.fill_lwe_mask_and_body_for_encryption(
```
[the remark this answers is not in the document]
IceTDrinker: as answered above the EncryptionRandomGenerator encapsulates two separate CSPRNG, one "publicly" seeded for the mask and one "privately" seeded for the noise

concrete-csprng/src/generators/implem/soft/generator.rs
```diff
@@ -17,16 +17,25 @@ impl Iterator for SoftwareChildrenIterator {
     fn next(&mut self) -> Option<Self::Item> {
         self.0.next().map(SoftwareRandomGenerator)
     }
+
+    // advance_by is unstable, the current recommendation is to implement nth to advance an iterator
+    // if the underlying generator has an efficient implementation this is better than the default
+    // nth which just calls next repeatedly
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
```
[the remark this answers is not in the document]
IceTDrinker: same as above

concrete-core/src/specification/engines/lwe_seeded_to_lwe_ciphertext_transmutation.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// A trait for engines transmuting LWE seeded ciphertexts into LWE ciphertexts.
+///
+/// # Semantics
+///
+/// This [pure](super#operation-semantics) operation moves the existing LWE seeded ciphertext into a
+/// LWE ciphertext.
+///
+/// # Formal Definition
```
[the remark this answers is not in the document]
IceTDrinker: thanks for the formal definition, I'll update it in the next round of patches!

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
IceTDrinker: Thanks as well, will integrate this definition

concrete-csprng/src/generators/implem/aesni/generator.rs
```diff
@@ -17,16 +17,25 @@ impl Iterator for AesniChildrenIterator {
     fn next(&mut self) -> Option<Self::Item> {
         self.0.next().map(AesniRandomGenerator)
     }
+
+    // advance_by is unstable, the current recommendation is to implement nth to advance an iterator
+    // if the underlying generator has an efficient implementation this is better than the default
+    // nth which just calls next repeatedly
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
```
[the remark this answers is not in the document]
IceTDrinker: though while revisiting the PR, it seems we will not use the shift that was introduced in the original PR.

I'll remove the nth code for now and we'll see if we end up needing it in the next seeded entities PRs

concrete-core/src/specification/engines/lwe_seeded_ciphertext_encryption.rs
IceTDrinker: done

IceTDrinker:
As discussed we will first wait for #55 to be merged, the quick seeding discussed will be done in a separate PR, issue can be found here: https://github.com/zama-ai/concrete-core-internal/issues/250

IceTDrinker:
another rebase to take into account the new changes from the current code base.

Transmutation engines are out for Transformation engines and the DefaultEngine now stores the CSPRNG-backed seeder introduced in https://github.com/zama-ai/concrete-core/pull/62 for use for seeded entities

concrete-csprng/src/generators/implem/aesni/generator.rs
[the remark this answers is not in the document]
damienligier: thanks for the details

concrete-core/src/backends/default/implementation/entities/lwe_seeded_ciphertext.rs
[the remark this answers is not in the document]
damienligier: @agnesLeroy i does not make much sense this restriction. could we change that?

concrete-core/src/backends/default/implementation/engines/mod.rs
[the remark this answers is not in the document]
damienligier: do you think we could have a better description in the comment?

concrete-core-fixture/src/generation/prototypes/lwe_seeded_ciphertext.rs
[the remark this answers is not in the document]
damienligier: @agnesLeroy here also i think we should not restrict to binary.

concrete-core/src/backends/default/implementation/engines/mod.rs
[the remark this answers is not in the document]
IceTDrinker: Of course, I'll indicate it has two CSPRNGs one publicly seeded for masks and one privately seeded for noise

concrete-core/src/backends/default/implementation/engines/mod.rs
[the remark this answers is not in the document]
IceTDrinker: updated

concrete-core/src/backends/default/implementation/engines/mod.rs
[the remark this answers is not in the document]
IceTDrinker: updated the comment

concrete-core/src/backends/default/implementation/entities/lwe_seeded_ciphertext.rs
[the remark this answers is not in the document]
agnesLeroy: @damienligier this is actually at the core of the V1 design, so it cannot be changed without an extensive rework of the code base. It is currently done this way because when we expose APIs to other languages on top of Concrete-core we have to explicit the integer representation and the key distribution. At least I don't see how we could do differently while being able to automate the API generation (which we haven't even done yet!). This is also a building block for the test & benchmark framework we implemented, that encompasses all backends and engines. It's maybe possible to do better, but requires some heavy design/implementation work.

If I'm not mistaken this question is also related to the issue we had with the tensor product operation. I haven't had time yet to describe the problem in more details in the issue (first draft here: https://github.com/zama-ai/concrete-core-internal/issues/242). @aPere3 is also working on a design issue to analyze the V1 design and explain why certain choices were made. As soon as we've written those issues I'll send you the link :slightly_smiling_face:

concrete-core/src/specification/engines/lwe_seeded_to_lwe_ciphertext_transmutation.rs
IceTDrinker: done

concrete-core/src/backends/default/implementation/engines/mod.rs
IceTDrinker: Done as we now use the Deterministic seeder that was added

agnesLeroy (approved):
Looks good to me @IceTDrinker! Thanks a lot for the contribution!

concrete-core/src/backends/default/implementation/entities/lwe_seeded_ciphertext.rs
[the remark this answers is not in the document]
damienligier: we should remove the key distribution for now.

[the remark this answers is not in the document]
agnesLeroy: I have opened a design issue for this question: https://github.com/zama-ai/concrete-core-internal/issues/258. This is just a draft for now. @damienligier could you maybe explain there why you find the use of the markers strange? Note that there is an open design issue for the tensor product (also a draft at this stage): https://github.com/zama-ai/concrete-core-internal/issues/242

damienligier (approved):

PR #26, opened 2022-05-25
feat(core): add an aesni backend for the key creations

agnesLeroy:
### Resolves

zama-ai/concrete-core-internal#170

### Description
This PR introduces an aesni backend to accelerate key generations.
WARNING: it is missing tests since we currently don't have tests for the key creations! There is a follow up issue for that. 
### Checklist 

* [ ] ~Tests for the changes have been added (for bug fixes / features)~

### DIFF
--- a/concrete-core/Cargo.toml
+++ b/concrete-core/Cargo.toml
@@ -46,6 +46,11 @@ backend_fftw = [
     "concrete-fftw"
 ]
 
+# An accelerated backend, using the `aesni` feature of x86_64 processors.
+backend_x86_64_aesni = [
+    "concrete-csprng/generator_x86_64_aesni"
+]
+
 seeder_unix = ["concrete-csprng/seeder_unix"]
 seeder_x86_64_rdseed = ["concrete-csprng/seeder_x86_64_rdseed"]
 parallel = ["rayon", "concrete-csprng/parallel"]
--- a/concrete-core/src/backends/aesni/implementation/engines/glwe_secret_key_creation.rs
+++ b/concrete-core/src/backends/aesni/implementation/engines/glwe_secret_key_creation.rs
@@ -1,0 +1,112 @@
+use concrete_commons::parameters::{GlweDimension, PolynomialSize};
+
+use crate::backends::aesni::implementation::engines::AesniEngine;
+use crate::commons::crypto::secret::GlweSecretKey as ImplGlweSecretKey;
+use crate::prelude::{GlweSecretKey32, GlweSecretKey64};
+use crate::specification::engines::{GlweSecretKeyCreationEngine, GlweSecretKeyCreationError};
+
+/// # Description:
+/// Implementation of [`GlweSecretKeyCreationEngine`] for [`AesniEngine`] that operates on
+/// 32 bits integers.
+impl GlweSecretKeyCreationEngine<GlweSecretKey32> for AesniEngine {
+    /// # Example:
+    /// ```
+    /// use concrete_commons::parameters::{GlweDimension, PolynomialSize};
+    /// use concrete_core::prelude::*;
+    /// # use std::error::Error;
+    ///
+    /// # fn main() -> Result<(), Box<dyn Error>> {
+    /// // DISCLAIMER: the parameters used here are only for test purpose, and are not secure.
+    /// let glwe_dimension = GlweDimension(2);
+    /// let polynomial_size = PolynomialSize(4);
+    ///
+    /// // Unix seeder must be given a secret input.
+    /// // Here we just give it 0, which is totally unsafe.
+    /// const UNSAFE_SECRET: u128 = 0;
+    /// let mut default_engine = DefaultEngine::new(Box::new(UnixSeeder::new(UNSAFE_SECRET)))?;
+    /// let mut aesni_engine = AesniEngine::new(Box::new(UnixSeeder::new(UNSAFE_SECRET)))?;
+    /// let glwe_secret_key: GlweSecretKey32 =
+    ///     aesni_engine.create_glwe_secret_key(glwe_dimension, polynomial_size)?;
+    /// #
+    /// assert_eq!(glwe_secret_key.glwe_dimension(), glwe_dimension);
+    /// assert_eq!(glwe_secret_key.polynomial_size(), polynomial_size);
+    ///
+    /// default_engine.destroy(glwe_secret_key)?;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn create_glwe_secret_key(
+        &mut self,
+        glwe_dimension: GlweDimension,
+        polynomial_size: PolynomialSize,
+    ) -> Result<GlweSecretKey32, GlweSecretKeyCreationError<Self::EngineError>> {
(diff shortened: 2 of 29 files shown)


concrete-core/Cargo.toml
```diff
@@ -46,6 +46,11 @@ backend_fftw = [
     "concrete-fftw"
 ]
 
+# An accelerated backend, using the `aesni` feature of x86_64 processors.
+backend_aesni = [
+    "concrete-csprng/generator_x86_64_aesni"
+]
```
IceTDrinker, line 52: we may want to keep the x86_64 name in the backend_aesni feature, wdyt ?

I know it makes it a bit longer but this way when enabling features it is crystal clear it requires an x86_64 CPU

concrete-core/src/backends/aesni/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! A pure-rust backend.
```
IceTDrinker, line 1: While this is true it's also accelerated (and therefore dependant on hardware features), could be worth having that reflected here ?

concrete-core/src/backends/aesni/implementation/engines/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl Display for AesniError {
+    fn fmt(&self, _f: &mut Formatter<'_>) -> std::fmt::Result {
+        match *self {}
+    }
+}
+
+impl Error for AesniError {}
+
+pub struct AesniEngine {
```
IceTDrinker, line 33: Is it something we want to expose in the C FFI down the line for efficiency reasons ? 😅

concrete-core/src/backends/aesni/implementation/engines/glwe_secret_key_creation.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// # fn main() -> Result<(), Box<dyn Error>> {
+    /// // DISCLAIMER: the parameters used here are only for test purpose, and are not secure.
+    /// let glwe_dimension = GlweDimension(2);
+    /// let polynomial_size = PolynomialSize(4);
+    ///
+    /// // Unix seeder must be given a secret input.
+    /// // Here we just give it 0, which is totally unsafe.
+    /// const UNSAFE_SECRET: u128 = 0;
+    /// let mut default_engine = DefaultEngine::new(Box::new(UnixSeeder::new(UNSAFE_SECRET)))?;
+    /// let mut aesni_engine = AesniEngine::new(Box::new(UnixSeeder::new(UNSAFE_SECRET)))?;
```
IceTDrinker, line 80: unrelated to the PR, but this is a pattern I find a bit clunky (needing a different engine to destroy an entity), I don't really know what can be done to make this a bit more consistent/less tiresome (having to create multiple engines) while keeping the current flat API

concrete-core/src/backends/aesni/implementation/engines/lwe_bootstrap_key_creation.rs
```diff
@@ -0,0 +1,200 @@
...
+        decomposition_level_count: DecompositionLevelCount,
+        noise: Variance,
+    ) -> LweBootstrapKey32 {
+        let mut key = ImplStandardBootstrapKey::allocate(
+            0,
+            output_key.0.key_size().to_glwe_size(),
+            output_key.0.polynomial_size(),
+            decomposition_level_count,
+            decomposition_base_log,
+            input_key.0.key_size(),
```
IceTDrinker, line 95: here as well (?)

concrete-core/src/backends/aesni/implementation/engines/lwe_bootstrap_key_creation.rs
```diff
@@ -0,0 +1,200 @@
...
+        noise: Variance,
+    ) -> LweBootstrapKey64 {
+        let mut key = ImplStandardBootstrapKey::allocate(
+            0,
+            output_key.0.key_size().to_glwe_size(),
+            output_key.0.polynomial_size(),
+            decomposition_level_count,
+            decomposition_base_log,
+            input_key.0.key_size(),
+        );
```
IceTDrinker, line 191: same here for the impl thing vs public API methods

IceTDrinker (commented):
Hello @agnesLeroy!

Thanks for the PR 🙂
Looks good overall, a few comments here and there and mostly a question about how to name the backend_aesni feature.

Cheers

concrete-core/Cargo.toml
agnesLeroy: Totally right thanks!

concrete-core/src/backends/aesni/implementation/engines/mod.rs
agnesLeroy: Yes absolutely :sweat_smile:

concrete-core/src/backends/aesni/implementation/engines/glwe_secret_key_creation.rs
agnesLeroy: We definitely need to think about this: we had no time until now :sweat_smile: A very rough first idea was to have a kind of compound engine but then depending on how you define it and your HW, we'll probably end up having to do some run time choices... that's as far as we got in the reflection (this is actually related to the backend support for higher level libraries).

concrete-core/src/backends/aesni/implementation/engines/lwe_bootstrap_key_creation.rs
```diff
@@ -0,0 +1,200 @@
...
+        input_key: &LweSecretKey32,
+        output_key: &GlweSecretKey32,
+        decomposition_base_log: DecompositionBaseLog,
+        decomposition_level_count: DecompositionLevelCount,
+        noise: Variance,
+    ) -> LweBootstrapKey32 {
+        let mut key = ImplStandardBootstrapKey::allocate(
+            0,
+            output_key.0.key_size().to_glwe_size(),
+            output_key.0.polynomial_size(),
```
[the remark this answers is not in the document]
agnesLeroy: I'll change it here :slightly_smiling_face: That's also what I did last time.

concrete-core/src/backends/aesni/implementation/engines/mod.rs
IceTDrinker: ok well not sure it's going to be in v0 🙃 because today I think the compiler is using the software version through the previous private implementations that were in concrete-core

What about a parallel engine using Aesni ? 😬 

My feeling is the parallel software engine should be faster than signle thread Aesni but I could be wrong

agnesLeroy: Yes I definitely need to add that, as well as all the other missing entry points for encryption! Thanks for pointing it out!

IceTDrinker (commented):
I'm wondering if the code duplication between the default engine and the aesni engine is worth it :|

IceTDrinker (commented):
Maybe worth thinking about what to do with the `#` signs in doc tests, their use does not seem consistent

agnesLeroy:
We're doing #40 instead

PR #29, opened 2022-05-30
test(ci): Format AWS action file to be usable with CI bot

soonum:
### Description

Format AWS tests action file to be usable by the Slab the CI-bot.

### Checklist 

* [x] Check for breaking changes and add them to commit message following the conventional commit [specification][conventional-breaking

### DIFF
--- a/.github/workflows/aws_tests_slab_beta.yml
+++ b/.github/workflows/aws_tests_slab_beta.yml
@@ -1,0 +1,52 @@
+# Compile and test project on an AWS instance
+name: AWS tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      instance_id:
+        description: 'AWS instance ID'
+        type: string
+      instance_image_id:
+        description: 'AWS instance AMI ID'
+        type: string
+      instance_type:
+        description: 'AWS EC2 instance product type'
+        type: string
+      runner_name:
+        description: 'Action runner name'
+        type: string
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  run-tests-linux:
+    concurrency:
+      group: ${{ github.head_ref }}
+      cancel-in-progress: true
+    name: Test code in EC2
+    runs-on: ${{ github.event.inputs.runner_name }}
+    strategy:
+      fail-fast: false
+      # explicit include-based build matrix, of known valid options
+      matrix:
+        include:
+          - os: ubuntu-20.04
+    steps:
+      - name: EC2 instance configuration used
+        run: |
+          echo "IDs: ${{ github.event.inputs.instance_id }}"
+          echo "AMI: ${{ github.event.inputs.instance_image_id }}"
+          echo "Type: ${{ github.event.inputs.instance_type }}"
+      - uses: actions/checkout@v2
+      - name: Set up home
+        run: |
+          echo "HOME=/home/ubuntu" >> "${GITHUB_ENV}"
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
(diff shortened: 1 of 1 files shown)


IceTDrinker (commented):
Can we keep both workflows for now to be able to debug and still use the existing one if we have issues ?

soonum:
You're right, we must have to keep the two workflow, moreover since Slab will be in beta for now.
I will create a new file then, what about `aws_tests_slab_beta.yml` ?

IceTDrinker:
> You're right, we must have to keep the two workflow, moreover since Slab will be in beta for now. I will create a new file then, what about `aws_tests_slab_beta.yml` ?

sounds good !

IceTDrinker (commented):
Hmm I see aws_tests.yml is gone (?)

There is a check for commit messages the format is:

```
          pattern: '^((feat|fix|chore|refactor|style|test|docs)\(\w+\)\:) .+$'
          flags: 'gs'
```

basically here you can probably use:

```
chore(ci): my commit message

commit body
```

We tend to force push on PRs for reviews for now

.github/workflows/aws_tests_slab_beta.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: EC2 instance configuration used
+        run: |
+          echo "IDs: ${{ github.event.inputs.instance_id }}"
+          echo "AMI: ${{ github.event.inputs.instance_image_id }}"
+          echo "Type: ${{ github.event.inputs.instance_type }}"
+      - uses: actions/checkout@v2
+      - name: Set up home
+        run: |
+          echo "HOME=/home/ubuntu" >> ${GITHUB_ENV}
```
IceTDrinker, line 45: missing double quotes around `${GITHUB_ENV}` to be safe 🙂

.github/workflows/aws_tests_slab_beta.yml
```diff
@@ -0,0 +1,52 @@
...
+      group: ${{ github.head_ref }}
+      cancel-in-progress: true
+    name: Test code in EC2
+    runs-on: ${{ github.event.inputs.runner_name }}
+    strategy:
+      fail-fast: false
+      # explicit include-based build matrix, of known valid options
+      matrix:
+        include:
+          - os: ubuntu-20.04
```
IceTDrinker, line 35: This may get out of sync as a runner's OS is defined by the AMI used for the run, don't know how to manage this.

I think it may already be a problem in our current aws_tests.yml

It's not really needed as far as I know as well as we don't use the matrix os info in the job, thoughts on this @agnesLeroy ?

IceTDrinker (commented):
one small thing :)

then I think you can squash your commits locally and force push the squash result with the quote thing to the PR, this way we'll only have a single commit 👍 

is it possible to test the bot already on this PR maybe ?

.github/workflows/aws_tests_slab_beta.yml
agnesLeroy: I think for now it's up to us not to mess up with the AMIs, I don't really see how we could do differently :thinking:

IceTDrinker: true but I'm questioning the very existance of this os entry, I don't think we need it 🙂

IceTDrinker (approved):
Thx @soonum !

For the os thing in the matrix don't bother, we'll see if we really need it but it's not linked to this PR's functionality 🙂

agnesLeroy:
@soonum just a question: why did you cancel the linux & mac builds?

PR #32, opened 2022-06-01
Update to the Optalysys backend

FlorentCLMichel:
This PR does three changes:

* change the imports to match the new crate structure (replacing imports from `backends::core` by imports from `commons` and re-implementing `Complex64` and `twiddles` in `backends::optalysys`);

* update the tests of the Optalysys backend;

* replace `AlignedVec`s by `Vec`s in the Optalysys backend, as a first step toward removing the `fftw` dependency for this backend;


agnesLeroy:
Hi @FlorentCLMichel ! Thank you for this PR :smile: I think you can reuse the twiddles file from the default backend by making it pub(crate) rather than copying it in the optalysys backend, no? Same for Complex64 maybe?

Why do you need to remove the Eq trait for the LWE secret keys? This is something we rely on in concrete-core and it may cause problems to remove it :thinking:

FlorentCLMichel:
Good points! 

For the twiddles and Complex64, if I am not mistaken they were defined in the `fftw` backend but not in the `default` one, so I copied the corresponding files in the `optalysys` backend to avoid having to load the `fftw` one. Happy to load them from `fftw` or from somewhere else if you think it is better, though! 

Concerning the `Eq` trait, I removed this requirement because `LweSecretKey32` and `LweSecretKey64` in the `default` backend did not seem to implement it (unless I missed it). I've just amended the PR to implement this trait and re-add the requirement.

FlorentCLMichel:
The `optalysys` backend should now use the same `Complex64` type and twiddles as the `fftw` one. (the corresponding file was moved to `commons/math/fft` so that both backends have access to it without making `private` folder public; thanks @agnesLeroy for the suggestion!) 

Thanks again @agnesLeroy for your helpful comments and help in Slack! Please let me know if anything else should be changed or/and done differently!

agnesLeroy:
> The `optalysys` backend should now use the same `Complex64` type and twiddles as the `fftw` one. (the corresponding file was moved to `commons/math/fft` so that both backends have access to it without making `private` folder public; thanks @agnesLeroy for the suggestion!)
> 
> Thanks again @agnesLeroy for your helpful comments and help in Slack! Please let me know if anything else should be changed or/and done differently!

It looks much better like this! Thanks!

concrete-core/Cargo.toml
```diff
@@ -23,10 +23,12 @@ serde_test = "1.0.125"
 [dependencies]
 concrete-fftw = { version = "=0.1.2", optional = true}
 concrete-commons = "=0.2.0"
-concrete-csprng = {version = "0.2", features = []}
+#concrete-csprng = {version = "0.2", features = []} 
+concrete-csprng = {path = "../concrete-csprng", features = []} # TO CHANGE
```
agnesLeroy, line 27: I think you maybe forgot to remove this change?

FlorentCLMichel: Oh, you're right! Thanks for pointing this out! Let me correct this now.

concrete-core/Cargo.toml
```diff
@@ -23,10 +23,11 @@ serde_test = "1.0.125"
 [dependencies]
 concrete-fftw = { version = "=0.1.2", optional = true}
 concrete-commons = "=0.2.0"
-concrete-csprng = {version = "0.2", features = []}
+concrete-csprng = {version = "0.2", features = []} 
 serde = { version = "1.0", optional = true }
 lazy_static = "1.4.0"
 rayon = { version = "1.5.0", optional = true }
+num = { version = "0.4.0", optional = true }
```
agnesLeroy, line 30: Here in order to exactly fit what's in concrete-fftw, it should be num-complex I think.

agnesLeroy (approved):

PR #39, opened 2022-06-08
Fixture: place the prototype key transmutation in dedicated traits

agnesLeroy:
### Description
 This PR changes the logic for prototyping GLWE & LWE secret keys via a transmutation. It is necessary to implement the tensor product fixture.

### Checklist 

### DIFF
--- a/concrete-core-fixture/src/generation/prototyping/lwe_secret_key.rs
+++ b/concrete-core-fixture/src/generation/prototyping/lwe_secret_key.rs
@@ -1,10 +1,14 @@
 use crate::generation::prototypes::{
-    LweSecretKeyPrototype, ProtoBinaryLweSecretKey32, ProtoBinaryLweSecretKey64,
+    LweSecretKeyPrototype, ProtoBinaryGlweSecretKey32, ProtoBinaryGlweSecretKey64,
+    ProtoBinaryLweSecretKey32, ProtoBinaryLweSecretKey64,
 };
+use crate::generation::prototyping::PrototypesGlweSecretKey;
 use crate::generation::{IntegerPrecision, Maker, Precision32, Precision64};
 use concrete_commons::parameters::LweDimension;
 use concrete_core::prelude::markers::{BinaryKeyDistribution, KeyDistributionMarker};
-use concrete_core::prelude::LweSecretKeyCreationEngine;
+use concrete_core::prelude::{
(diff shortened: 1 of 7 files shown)


concrete-core-fixture/src/generation/prototyping/lwe_secret_key.rs
```diff
@@ -44,3 +48,45 @@ impl PrototypesLweSecretKey<Precision64, BinaryKeyDistribution> for Maker {
...
+    ) -> Self::GlweSecretKeyProto {
+        ProtoBinaryGlweSecretKey32(
+            self.default_engine
+                .transmute_lwe_secret_key_to_glwe_secret_key(lwe_key.0.to_owned(), polynomial_size)
+                .unwrap(),
+        )
+    }
+}
+
+impl TransmutesLweSecretKeyPrototype<Precision64, BinaryKeyDistribution> for Maker {
```
IceTDrinker, line 80: Same remark for all Transmutes trait you added:

either make the name more explicit to know towards what type we transmute e.g.

`TransmutesLweToGlweSecretKeyPrototype`

or make it generic over the output type and add said output type in the impls blocks

IceTDrinker (commented):
Hello @agnesLeroy!

Thanks for the PR (and sorry for the annoyance, I remember adding that transmutation thing a while back 😬 ) the only remark I have is with naming or making the trait generic over the return type so that it's more explicit.

Other than that looks good!

Thanks

concrete-core-fixture/src/generation/prototyping/lwe_secret_key.rs
IceTDrinker: unsure for the generic thing, it could bring more problems than it's worth, up to you!

Naming should be fine for now (as always let's prepare what we need 🙂 )

agnesLeroy: Yes, trying to make the output type generic seems like a lot of trouble. I've just changed the names :sweat_smile: If we get more transmutation options later we could reconsider this when the time comes.

IceTDrinker: yes exactly !

IceTDrinker (approved):
Thanks a lot!

PR #49, opened 2022-06-14
chore(concrete_core): raise the `DefaultParallelEngine` in super module

aPere3:
### Resolves:

closes zama-ai/concrete-core-internal#227

### Description

When the DefaultParallelEngine was added, it was inside of a parallel module, and as such, it is not at the same level in the modules as the DefaultEngine.

This commit solves it.

### Checklist 

* ~[ ] Tests for the changes have been added (for bug fixes / features)~
* ~[ ] Docs have been added / updated (for bug fixes / features)~
* ~[ ] The draft release description has been updated~

### DIFF
--- a/concrete-core/src/backends/default/implementation/engines/mod.rs
+++ b/concrete-core/src/backends/default/implementation/engines/mod.rs
@@ -62,7 +62,7 @@ impl AbstractEngine for DefaultEngine {
 }
 
 #[cfg(feature = "parallel")]
-pub mod parallel {
+mod parallel {
     use std::error::Error;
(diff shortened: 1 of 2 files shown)


agnesLeroy (approved):
Thanks for this Alex! :slightly_smiling_face:

agnesLeroy:
Is it normal that the PR is still in draft? The tests fail but it's only the uniform test in csprng that sometimes does. The clippy failure should be solved when rebasing.

aPere3:
Actually, we will rework slightly the organization of the different engines in the PR #55 .

PR #55, opened 2022-06-15
Feature/serialization

aPere3:
### Resolves: 

closes zama-ai/concrete-core-internal#95
closes zama-ai/concrete-core-internal#72
closes zama-ai/concrete-core-internal#167
closes zama-ai/concrete-core-internal#222

### Description

This commit adds support for serialization in `concrete-core`. Two engine traits were added to the `specification` module. This trait was then implemented in both the `default` and the `fftw` backend, for all existing entities.

In addition, an early support for serialization versioning was added. All entities are currently tagged with the `V0` version, which is then checked at deserialization.

This commit updates `concrete_ffi` to use the newly introduced serialization support in `concrete_core`.

### Checklist 


concrete-ffi/Cargo.toml
```diff
@@ -28,14 +28,15 @@ doc = [
     "parallel",
     "seeder_unix",
     "seeder_x86_64_rdseed",
-    "serde_serialize",
```
IceTDrinker, line 31: we may want to have the serialization features when building docs ?

I think I followed the pattern from somewhere but can't find it anymore, maybe this should be removed, I'm not sure

concrete-core/src/specification/engines/entity_deserialization.rs
```diff
@@ -0,0 +1,33 @@
...
+    UnsupportedVersion => "The version used to serialize the entity is not supported."
+}
+
+/// A trait for engines deserializing entities.
+///
+/// # Semantics
+///
+/// This [pure](super#operation-semantics) operation generates an entity containing the
+/// deserialization of the `serialized` type.
+pub trait EntityDeserializationEngine<Entity, Serialized>: AbstractEngine
```
IceTDrinker, line 16: type parameters are backwards I would say

should Entity have a trait bound of AbastractEntity ?

concrete-core/src/backends/default/implementation/engines/entity_serialization.rs
```diff
@@ -0,0 +1,903 @@
...
+    ) -> Result<Vec<u8>, EntitySerializationError<Self::EngineError>> {
+        #[derive(Serialize)]
+        struct SerializableCleartext64<'a> {
+            version: Cleartext64Version,
+            inner: &'a ImplCleartext<u64>,
+        }
+        let serializable = SerializableCleartext64 {
+            version: Cleartext64Version::V0,
+            inner: &entity.0,
+        };
```
IceTDrinker, line 78: it feels tricky to have versioning here while the content of a Cleartext64 is defined elsewhere and not tied to the version 🤔 cleartexts are simple structs but for more elaborate ones I think the loose coupling might create issues down the line

concrete-core/src/backends/default/implementation/engines/entity_deserialization.rs
```diff
@@ -0,0 +1,877 @@
...
+    LweKeyswitchKey64Version, LweSecretKey32, LweSecretKey32Version, LweSecretKey64,
+    LweSecretKey64Version, PackingKeyswitchKey32, PackingKeyswitchKey32Version,
+    PackingKeyswitchKey64, PackingKeyswitchKey64Version, Plaintext32, Plaintext32Version,
+    Plaintext64, Plaintext64Version, PlaintextVector32, PlaintextVector32Version,
+    PlaintextVector64, PlaintextVector64Version,
+};
+use concrete_commons::key_kinds::BinaryKeyKind;
+use serde::Deserialize;
+
+impl EntityDeserializationEngine<Cleartext32, &[u8]> for DefaultEngine {
```
IceTDrinker, line 39: is there a way to perhaps proc macro serialization/deserialization functions, looks like it's always the same code

concrete-core/src/backends/default/implementation/engines/mod.rs
```diff
@@ -20,11 +20,30 @@ use crate::specification::engines::AbstractEngine;
...
+            DefaultError::Serialization(bincode_error) => {
+                write!(f, "Failed to serialize entity: {}", bincode_error)
+            }
+            #[cfg(feature = "backend_default_serialization")]
+            DefaultError::Deserialization(bincode_error) => {
+                write!(f, "Failed to deserialize entity: {}", bincode_error)
+            }
+            // For some reason this is needed to avoid a compiler error when
+            // `backend_default_serialization` is not active.
+            _ => unreachable!(),
```
IceTDrinker, line 45: I'm guessing matching on an enum without variant is a bit of a weird case 🙂

concrete-core/src/backends/default/implementation/entities/cleartext.rs
```diff
@@ -10,10 +12,26 @@ impl AbstractEntity for Cleartext32 {
 }
 impl CleartextEntity for Cleartext32 {}
 
+#[cfg(feature = "backend_default_serialization")]
+#[derive(Serialize, Deserialize)]
+pub(crate) enum Cleartext32Version {
+    V0,
+    #[serde(other)]
+    Unsupported,
+}
```
IceTDrinker, line 21: maybe we need to add a `#[repr(u8)]` (or bigger bitwidth) to make sure values associated to variants stay consistent ? I don't know what the rust compiler does for enuls

concrete-ffi/Cargo.toml
aPere3: Yes, sorry, I forgot to add the new features :laughing: !

Done !

concrete-core/src/specification/engines/entity_deserialization.rs
aPere3: You mean that the `Serialized` type being the input and the `Entity` being the output, they should appear in reverse order ?

If so, I think you are right ...

IceTDrinker: yep it was about the order 🙂

concrete-core/src/backends/default/implementation/engines/entity_serialization.rs
aPere3: See #248 for the decision on this matter.

IceTDrinker:
as you are also fixing the C FFI you will be closing this issue as well: https://github.com/zama-ai/concrete-core-internal/issues/167

concrete-core/src/backends/default/implementation/engines/entity_serialization.rs
```diff
@@ -0,0 +1,903 @@
...
+        bincode::serialize(&serializable)
+            .map_err(DefaultError::Serialization)
+            .map_err(EntitySerializationError::Engine)
+    }
+
+    unsafe fn serialize_unchecked(&mut self, entity: &Cleartext64) -> Vec<u8> {
+        self.serialize(entity).unwrap()
+    }
```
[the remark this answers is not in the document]
aPere3: Yeah, it also felt weird to me at first, that it was working in reverse :confused: 

Actually, some `unchecked` functions may panic yes. So in the end I did not really had a choice, because bincode may produce an error that I must be able to catch. For that I need the bincode call to happen in the checked entry point otherwise I won't be able to capture it. And since the `unchecked` variant is supposed to return the `Vec` directly, well. I ended up doing it like that ...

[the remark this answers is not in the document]
agnesLeroy: In this case shouldn't we propose only the checked variant?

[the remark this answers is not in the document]
aPere3: Do you mean in the spec ? I don´t think so. We have unchecked entry points everywhere else, I don't see a reason to change that. 

Also maybe things would be different with a different engine ?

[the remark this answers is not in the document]
agnesLeroy: Sorry I wasn't clear: are we compelled to implement the two variants in the default backend? We could keep the two in the spec but implement only the checked variant in the default backend?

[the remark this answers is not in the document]
agnesLeroy: Oh we'll need the unchecked variant for the c_ffi right?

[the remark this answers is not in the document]
aPere3: Exactly :)

concrete-core/src/backends/default/implementation/engines/entity_deserialization.rs
aPere3: I have to write something about code-duplication in `concrete-core` and the reasons that made us go the way things are. Long story short, there are two main topics:
+ It is sometimes difficult to craft an abstraction that is, in practice, cheaper to maintain than duplicated code.
+ In our case here, we can only go the macro way (procedural or not).
    -> If we use macro-by-example, we end up with huge blocks of code with no ide support
    -> If we use proc-macros, we end up with an abstraction in a separate place, which uses completely different patterns than the one usually found in the codebase. I believe that it is way more difficult for newcomers to work with.
    
For now, we decided that it was simpler, faster, and clearer for newcomers to just duplicate things. 

That being said, we will have to have this discussion at some point in the future, because at some point it will stop to be  _simpler, faster, clearer_. But I believe, we will have to have this discussion for a larger scope than just this pr :) 

For this reason, I would prefer to leave it that way, and maybe create an issue about that in  the backlog ?

concrete-core/src/backends/default/implementation/engines/entity_serialization.rs
```diff
@@ -0,0 +1,903 @@
...
+        entity: &Cleartext32,
+    ) -> Result<Vec<u8>, EntitySerializationError<Self::EngineError>> {
+        #[derive(Serialize)]
+        struct SerializableCleartext32<'a> {
+            version: Cleartext32Version,
+            inner: &'a ImplCleartext<u32>,
+        }
+        let serializable = SerializableCleartext32 {
+            version: Cleartext32Version::V0,
+            inner: &entity.0,
```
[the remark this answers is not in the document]
aPere3: No reasons ! :laughing: I don't know why I did it that way, but like you said it does not change anything. Unless you see a reason to change that, I think we can leave it that way ?

concrete-core/src/backends/default/implementation/engines/mod.rs
aPere3: That does not exist anymore ;)

concrete-core/src/backends/default/implementation/engines/entity_deserialization.rs
IceTDrinker: No worries I understand the general reasons 🙂 

Fine for creating an issue, should you do it ?

concrete-core/src/backends/default/implementation/engines/entity_serialization.rs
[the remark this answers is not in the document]
IceTDrinker: I would rather not serialize something that comes from the private API contained in the .0, WDYT ?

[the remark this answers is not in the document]
agnesLeroy: Maybe in the future public entities may be more than a wrapper to a private struct? In that case we should serialize the entity itself and not only the content of `.0`?

concrete-core/src/backends/default/implementation/entities/cleartext.rs
aPere3: Hmmm interesting point. That being said, by reading the serde documentation, I understand that variants are always represented by a string (which should work consistently). From my understanding, `#[repr(u8)]` would not have impact on how `serde` represent things.

concrete-core/src/specification/engines/entity_deserialization.rs
aPere3: Ok I'll fix that :wink: Thanks for noticing !

concrete-core/src/backends/default/implementation/entities/cleartext.rs
IceTDrinker: ok good to know, if it's a String we should be good then!

concrete-core/src/backends/default/implementation/engines/entity_serialization.rs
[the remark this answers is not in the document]
aPere3: You are right ! I'll change that !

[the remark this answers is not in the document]
aPere3: Hmm now I remember why I did it this way haha !

It allows not to expose `Serialize` and `Deserialize` at the level of the entities. I found this was better, as we want people to go through the engine operators, rather than implement their own serialization. Also, the private implementation kind of leak through with the serialize trait.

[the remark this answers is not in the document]
aPere3: Se we went for the current solution in the end. See zama-ai/concrete-core-internal#248

concrete-core-ffi/tests/test_lwe_bootstrap_w_bsk_serde.c
```diff
@@ -54,6 +54,11 @@ void bootstrap_view_buffers_test(void) {
   int fftw_engine_ok = new_fftw_engine(&fftw_engine);
   assert(fftw_engine_ok == 0);
 
+  FftwSerializationEngine *fftw_serialization_engine = NULL;
```
IceTDrinker, line 57: I think you forget to destroy the serilialization engine in the tests

concrete-core-ffi/tests/test_lwe_encrypt_keyswitch_w_ksk_serde_decrypt.c
```diff
@@ -15,6 +15,11 @@ void keyswitch_view_buffers_test(void) {
   assert(default_engine_ok == 0);
   assert(engine != NULL);
 
+  DefaultSerializationEngine *default_serialization_engine = NULL;
```
IceTDrinker, line 18: Same here, serialization engine is leaking

concrete-core-ffi/tests/test_lwe_encrypt_w_sk_serde_decrypt.c
```diff
@@ -15,6 +15,11 @@ void encrypt_decrypt_view_buffers_test(void) {
   assert(default_engine_ok == 0);
   assert(engine != NULL);
 
+  DefaultSerializationEngine *default_serialization_engine = NULL;
```
IceTDrinker, line 18: Ditto

concrete-core/Cargo.toml
```diff
@@ -35,20 +36,38 @@ doc = []
...
+# Enables the parallel engine in default backend.
+backend_default_parallel = ["__commons_parallel"]
+
 # Enable the x86_64 specific accelerated implementation of the random generator for the default
 # backend
 backend_default_generator_x86_64_aesni = ["concrete-csprng/generator_x86_64_aesni"]
 
+# Enable the serialization engine in the default backend.
+backend_default_serialization = [
+    "bincode",
```
IceTDrinker, line 48: I'm guessing bincode cannot be put in the commons part as it uses a specific binary serialization format ?

concrete-core/src/backends/default/implementation/engines/default_engine/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+
+#[cfg(feature = "backend_default_generator_x86_64_aesni")]
+type ActivatedRandomGenerator = concrete_csprng::generators::AesniRandomGenerator;
+#[cfg(not(feature = "backend_default_generator_x86_64_aesni"))]
+type ActivatedRandomGenerator = concrete_csprng::generators::SoftwareRandomGenerator;
+
+pub struct DefaultEngine {
+    secret_generator: ImplSecretRandomGenerator<ActivatedRandomGenerator>,
+    encryption_generator: ImplEncryptionRandomGenerator<ActivatedRandomGenerator>,
+}
```
IceTDrinker, line 36: with the PR on CSPRNG backed Seeder merged this will need a rebase and conflict resolution

concrete-core/src/backends/default/implementation/engines/default_parallel_engine/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+impl AbstractEngineSeal for DefaultParallelEngine {}
+
+impl AbstractEngine for DefaultParallelEngine {
+    type EngineError = DefaultParallelError;
+
+    type Parameters = Box<dyn Seeder>;
+
+    fn new(mut parameters: Self::Parameters) -> Result<Self, Self::EngineError> {
+        Ok(DefaultParallelEngine {
+            encryption_generator: ImplEncryptionRandomGenerator::new(
```
IceTDrinker, line 50: will also need to update the code to use the deterministic Seeder

concrete-core/src/backends/default/implementation/engines/default_engine/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl AbstractEngine for DefaultEngine {
+    type EngineError = DefaultError;
+
+    type Parameters = Box<dyn Seeder>;
+
+    fn new(mut parameters: Self::Parameters) -> Result<Self, Self::EngineError> {
+        Ok(DefaultEngine {
+            secret_generator: ImplSecretRandomGenerator::new(parameters.seed()),
+            encryption_generator: ImplEncryptionRandomGenerator::new(
```
IceTDrinker, line 48: here as well

concrete-core/src/backends/fftw/private/crypto/ggsw/fourier.rs
```diff
@@ -18,11 +18,11 @@ use crate::commons::crypto::glwe::{GlweCiphertext, GlweList};
 use crate::commons::math::decomposition::{DecompositionLevel, SignedDecomposer};
 use crate::commons::math::polynomial::Polynomial;
 use crate::commons::math::torus::UnsignedTorus;
-#[cfg(feature = "serde_serialize")]
+#[cfg(feature = "backend_fftw_serialization")]
 use serde::{Deserialize, Serialize};
```
IceTDrinker, line 22: isn't serde_serialize in the commons one ?

concrete-core/src/backends/fftw/private/crypto/ggsw/fourier.rs
```diff
@@ -451,7 +451,7 @@ impl<Cont, Scalar> FourierGgswCiphertext<Cont, Scalar> {
     ///     .all(|a| *a == Complex64::new(0., 0.)));
     /// assert_eq!(ggsw.level_matrix_iter_mut().count(), 3);
     /// ```
-    #[cfg(feature = "parallel")]
+    #[cfg(feature = "__commons_parallel")]
```
IceTDrinker, line 454: isn't the feature name backend_fftw_serialization ?

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_deserialization.rs
```diff
@@ -0,0 +1,1938 @@
...
+    /// let noise = Variance(2_f64.powf(-25.));
+    ///
+    /// // Unix seeder must be given a secret input.
+    /// // Here we just give it 0, which is totally unsafe.
+    /// const UNSAFE_SECRET: u128 = 0;
+    /// let mut engine = DefaultEngine::new(Box::new(UnixSeeder::new(UNSAFE_SECRET)))?;
+    /// let input_key: LweSecretKey32 = engine.create_lwe_secret_key(input_lwe_dimension)?;
+    /// let output_key: LweSecretKey32 = engine.create_lwe_secret_key(output_lwe_dimension)?;
+    ///
+    /// let packing_keyswitch_key = engine.create_lwe_keyswitch_key(
```
IceTDrinker, line 1569: does this really create a packing keyswitch ? I see a `PackingKeyswitchKeyCreationEngine` with a function called `create_packing_keyswitch_key`

.github/workflows/ffi_github_release.yml
```diff
@@ -1,4 +1,4 @@
-name: concrete-core-ffi GitHub release
+name: concrete-ffi GitHub release
```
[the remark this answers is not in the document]
aPere3: Done.

.github/workflows/ffi_github_release.yml
```diff
@@ -27,11 +27,12 @@ jobs:
           cargo build --release \
           --features backend_default \
           --features backend_default_generator_x86_64_aesni \
+          --features backend_default_serialization \
           --features backend_fftw \
+          --features backend_fftw_serialization \
           --features parallel \
```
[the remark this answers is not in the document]
aPere3: Fixed

concrete-core-ffi/src/backends/default/engines/mod.rs
```diff
@@ -77,7 +77,7 @@ pub unsafe extern "C" fn destroy_default_engine_unchecked(engine: *mut DefaultEn
 ///
 /// This function is [checked](crate#safety-checked-and-unchecked-functions).
 #[no_mangle]
-#[cfg(feature = "parallel")]
+#[cfg(feature = "backend_default_parallel")]
```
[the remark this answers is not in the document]
aPere3: Not sure about what is tedious :laughing: Could you detail ?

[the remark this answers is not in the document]
IceTDrinker: Here all functions for parallel engine are feature gated, I think you have an additional level of file hierarchy in concrete-core could be worth reproducing here to avoid having all the cfgs ?

concrete-core-ffi/src/backends/default/engines/mod.rs
```diff
@@ -146,7 +146,77 @@ pub unsafe extern "C" fn destroy_default_parallel_engine_unchecked(
...
+) -> c_int {
+    catch_panic(|| {
+        check_ptr_is_non_null_and_aligned(result).unwrap();
+
+        // First fill the result with a null ptr so that if we fail and the return code is not
+        // checked, then any access to the result pointer will segfault (mimics malloc on failure)
+        *result = std::ptr::null_mut();
+
+        let heap_allocated_default_engine = Box::new(DefaultSerializationEngine::new(()).unwrap());
+        *result = Box::into_raw(heap_allocated_default_engine);
```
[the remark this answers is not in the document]
aPere3: Done !

concrete-core-ffi/src/backends/default/engines/entity_deserialization.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::utils::{
+    catch_panic, check_ptr_is_non_null_and_aligned, engine_error_as_readable_string,
+    get_mut_checked,
+};
+use concrete_core::prelude::{
+    DefaultSerializationEngine, EntityDeserializationEngine, LweKeyswitchKey64, LweSecretKey64,
+};
+use std::os::raw::c_int;
+
+/// Deserializes a `LweSecretKey64`.
```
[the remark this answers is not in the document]
aPere3: Done !

concrete-core-ffi/src/backends/default/engines/entity_deserialization.rs
```diff
@@ -0,0 +1,112 @@
...
+        *result = std::ptr::null_mut();
+
+        let engine = get_mut_checked(engine).unwrap();
+
+        let secret_key: LweSecretKey64 = engine
+            .deserialize(buffer.into())
+            .or_else(engine_error_as_readable_string)
+            .unwrap();
+
+        *result = Box::into_raw(Box::new(secret_key));
```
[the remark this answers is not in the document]
aPere3: Hmmm how would you do it then ? Because I think I reused what existed before in the serialization functions :confused:

concrete-core-ffi/tests/test_lwe_bootstrap_w_bsk_serde.c
aPere3: Yes :/ Fixed !

concrete-core/Cargo.toml
aPere3: you mean in the `__comons_serialization` feature ?

IceTDrinker: yes

concrete-core/src/backends/default/implementation/engines/default_engine/mod.rs
aPere3: Yes, I don't understand why we had nothing, because I rebased on your branch yesterday, but anyway. This is fixed now !

concrete-core/src/backends/default/implementation/engines/default_parallel_engine/mod.rs
aPere3: Done!

concrete-core/src/backends/default/implementation/engines/default_engine/mod.rs
aPere3: Done !

concrete-core-ffi/src/backends/default/engines/mod.rs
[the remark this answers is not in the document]
IceTDrinker: ok then I'll create an issue so we don't forget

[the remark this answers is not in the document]
IceTDrinker: https://github.com/zama-ai/concrete-core-internal/issues/255

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_deserialization.rs
aPere3: Hmm good catch, I did not even notice :cry: 

This snippet comes from:
https://github.com/zama-ai/concrete-core/blob/d338e8d587266c3b58ef0a67f1db84dc9363b549/concrete-core/src/backends/default/implementation/engines/packing_keyswitch_key_creation.rs#L20-L69

@agnesLeroy

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_serialization.rs
```diff
@@ -0,0 +1,2264 @@
...
+    /// engine.destroy(plaintext_vector)?;
+    /// engine.destroy(view)?;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    fn serialize(
+        &mut self,
+        entity: &GlweCiphertextView32<'b>,
+    ) -> Result<Vec<u8>, EntitySerializationError<Self::EngineError>> {
```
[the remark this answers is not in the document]
aPere3: Hmm that would be a problem for fixture, that is for sure. Btw, I did not added a fixture for those operators
@agnesLeroy , do you think we should add fixtures ?

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_deserialization.rs
IceTDrinker: Not to worry, that's why code duplication is something I talk about a lot 😄 

Issue for fixing the doc strings : https://github.com/zama-ai/concrete-core-internal/issues/256

Our issues now don't fit in a byte anymore 😛

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_serialization.rs
```diff
@@ -0,0 +1,2264 @@
...
+    /// const UNSAFE_SECRET: u128 = 0;
+    /// let mut engine = DefaultEngine::new(Box::new(UnixSeeder::new(UNSAFE_SECRET)))?;
+    /// let plaintext_vector: PlaintextVector64 = engine.create_plaintext_vector(&input)?;
+    ///
+    /// let mut serialization_engine = DefaultSerializationEngine::new(())?;
+    /// let serialized = serialization_engine.serialize(&plaintext_vector)?;
+    /// let recovered = serialization_engine.deserialize(serialized.as_slice())?;
+    /// assert_eq!(plaintext_vector, recovered);
+    ///
+    /// engine.destroy(plaintext_vector)?;
```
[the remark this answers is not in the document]
aPere3: Serialized does not need destroying (its a vec), but recovered does, you are right !

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_serialization.rs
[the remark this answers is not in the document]
IceTDrinker: I feel like views should not be part of the serialization mechanism, or perhaps they could only be serialized but deserialize to an actual LweCiphertext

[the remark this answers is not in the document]
aPere3: Yes, I believe this is what I did here. The `EntitySerializationEngine` trait is implemented for views, but not `EntityDeserializationEngine`.

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_deserialization.rs
aPere3: Well in practice here, it is the duplication that allowed us to find this guy :)

aPere3: Thanks for the issue, I close !

concrete-core/Cargo.toml
aPere3: Well, `bincode` is only used in the backend themselves, because it is where we choose to serialize to binary so I wouldn't put it in the commons dependencies.

concrete-core/src/backends/fftw/private/crypto/ggsw/fourier.rs
aPere3: I don't understand the question here :laughing:

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_deserialization.rs
IceTDrinker: > Well in practice here, it is the duplication that allowed us to find this guy :)

If the doc string surface was smaller (through less duplication) we would have less issues like this is what I meant 😅

concrete-core/src/backends/fftw/private/crypto/ggsw/fourier.rs
aPere3: I don't understand the question here :laughing:

concrete-core/Cargo.toml
IceTDrinker: yep that's what I was thinking in my first message, all good

concrete-core/src/backends/fftw/private/crypto/ggsw/fourier.rs
aPere3: Sorry, but I still don't understand :laughing: 

There is no `serde_derive` feature anymore now. What we have instead is:
+ a `__commons_serialization` feature, which include codeblocks in the `src/commons` module.
+ a set of `backends_*_serialization` which include codeblocks in the backend concerned.

The codeblock we are looking at is not in the `commons` module, but is in the private part of the `fftw` backend. I think this is why I did it this way. Is there anything wrong in the way it is done ? What would you do instead ?

IceTDrinker: My current thinking:

if the required dependency or feature is present in __commons_serialization then we should use that to feature gate as it could silently break if we use the super feature which does not directly activate the required dependency or feature.

It's debatable as you point out, we can still use the `backends_*_serialization` to enable the code in specific backends.

We can have a quick chat it's a bit hard to describe in writing sorry 😅

concrete-core/src/backends/default/implementation/engines/default_serialization_engine/entity_serialization.rs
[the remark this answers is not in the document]
agnesLeroy: @aPere3 I don't think we want fixtures no, we'll need thorough serialization tests instead.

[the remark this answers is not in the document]
IceTDrinker: Maybe we want to serialize views as an owning ciphertext ? copying the slice to a vec so that it can be deserialized ?

Because as is it feels like this serialization implementation for views is not super useful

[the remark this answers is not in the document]
IceTDrinker: I did not understand it's good like this !

IceTDrinker (approved):
Thanks for the hard work !

Let's serialize stuff 😄

agnesLeroy (approved):

PR #71, opened 2022-06-27
Feature/seeded glwe

IceTDrinker:
### Resolves:

closes https://github.com/zama-ai/concrete-core-internal/issues/260

### Description

GlweSeeded with facilities for compression only

Missing formal defs

### Checklist 


IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
Note that this PR is based on main and not the KSK PR to avoid having 40 commits while we wait for PRs to merge

concrete-core/src/commons/crypto/glwe/mod.rs
```diff
@@ -1,13 +1,15 @@
 //! GLWE encryption scheme
 
-pub use body::*;
-pub use ciphertext::*;
-pub use keyswitch::*;
-pub use list::*;
-pub use mask::*;
-
```
agnesLeroy, line 8: Maybe we can avoid exchanging the two sections here?

concrete-core/src/commons/crypto/glwe/seeded_ciphertext.rs
```diff
@@ -0,0 +1,361 @@
...
+pub struct GlweSeededCiphertext<Cont> {
+    tensor: Tensor<Cont>,
+    glwe_dimension: GlweDimension,
+    compression_seed: CompressionSeed,
+}
+
+tensor_traits!(GlweSeededCiphertext);
+
+impl<Scalar> GlweSeededCiphertext<Vec<Scalar>> {
+    /// Allocates a new GLWE seeded ciphertext, whose body and masks coefficients are all `value`.
```
agnesLeroy, line 27: Here the comment shouldn't mention the mask coefficient since we're only storing the body.

agnesLeroy: Maybe adding a little explanation about this here just like for the seeded ksk would be good actually.

concrete-core/src/commons/crypto/glwe/seeded_ciphertext.rs
```diff
@@ -0,0 +1,361 @@
...
+    ///
+    /// let glwe_seeded_ciphertext = GlweSeededCiphertext::<Vec<u8>>::allocate(
+    ///     polynomial_size,
+    ///     glwe_dimension,
+    ///     compression_seed,
+    /// );
+    ///
+    /// assert_eq!(glwe_seeded_ciphertext.size(), glwe_dimension.to_glwe_size());
+    /// ```
+    pub fn size(&self) -> GlweSize {
```
agnesLeroy, line 139: Here we should name it glwe_size to be consistent with our new naming convention

concrete-core/src/commons/crypto/glwe/seeded_ciphertext.rs
```diff
@@ -0,0 +1,361 @@
...
+    ///
+    /// let glwe_seeded_ciphertext = GlweSeededCiphertext::<Vec<u8>>::allocate(
+    ///     polynomial_size,
+    ///     glwe_dimension,
+    ///     compression_seed,
+    /// );
+    ///
+    /// assert_eq!(glwe_seeded_ciphertext.mask_size(), glwe_dimension);
+    /// ```
+    pub fn mask_size(&self) -> GlweDimension {
```
agnesLeroy, line 164: Do we need mask_size actually, since we already have glwe_size? I don't think we should keep it :)

concrete-core/src/commons/crypto/glwe/seeded_ciphertext.rs
```diff
@@ -0,0 +1,361 @@
...
+    ///     glwe_dimension,
+    ///     compression_seed,
+    /// );
+    ///
+    /// assert_eq!(
+    ///     glwe_seeded_ciphertext.get_compression_seed(),
+    ///     compression_seed
+    /// );
+    /// ```
+    pub fn get_compression_seed(&self) -> CompressionSeed {
```
agnesLeroy, line 218: in the other names we haven't written 'get', so we could just name it compression_seed.

agnesLeroy (commented):
Hello! Just some comments on this but overall it looks good! :)

concrete-core/src/specification/engines/glwe_seeded_ciphertext_to_glwe_ciphertext_transformation.rs
```diff
@@ -0,0 +1,47 @@
...
+/// # Formal Definition
+///
+/// ## GLWE seeded ciphertext to GLWE ciphertext transformation
+/// TODO
+pub trait GlweSeededCiphertextToGlweCiphertextTransformationEngine<
+    InputCiphertext,
+    OutputCiphertext,
+>: AbstractEngine where
+    InputCiphertext: GlweSeededCiphertextEntity,
+    OutputCiphertext: GlweCiphertextEntity,
```
aPere3, line 26: Here there should be a bound on key kinds. I was going to say that since we are going to remove them, we can drop it, but actually, maybe it would be better to keep follow this route until the commits that remove them ?

IceTDrinker: done

concrete-core/src/commons/crypto/glwe/mod.rs
IceTDrinker: well it makes it match the rest :| and it feels more logical to me to define the mod then use, though rust seems to not care about statements order

concrete-core/src/commons/crypto/glwe/seeded_ciphertext.rs
IceTDrinker: actually it matches the other glwe stuff, so I would leave as is

concrete-core/src/commons/crypto/glwe/seeded_ciphertext.rs
IceTDrinker: actually it matches the other glwe stuff, so I would leave as is

PR #80, opened 2022-07-01
fix(cuda): fix in cuda entities

agnesLeroy:
### Resolves: zama-ai/concrete-core-internal#286
(Also, use generic checks in the engines)

### Description
Deriving Clone, PartialEq and Eq onto Cuda entities is introducing a flaw in the cuda backend API. This PR fixes it.

### Checklist 

* [ ] Check for breaking changes and add them to commit message following the conventional commit [specification][conventional-breaking]


IceTDrinker (approved):
Thanks for the fix !

Does it impact fixtures or is it alright like this ?

agnesLeroy:
> Thanks for the fix !
> 
> Does it impact fixtures or is it alright like this ?

Tests will let me know :sweat_smile:

concrete-core/src/backends/cuda/implementation/engines/lwe_ciphertext_discarding_bootstrap.rs
```diff
@@ -126,22 +126,10 @@ impl
...
-        }
-        if output.lwe_dimension() != bsk.output_lwe_dimension() {
-            return Err(LweCiphertextDiscardingBootstrapError::OutputLweDimensionMismatch);
-        }
+        LweCiphertextDiscardingBootstrapError::perform_generic_checks(output, input, acc, bsk)?;
         let poly_size = bsk.0.polynomial_size().0;
         check_poly_size!(poly_size);
-        unsafe { self.discard_bootstrap_lwe_ciphertext_unchecked(output, input, acc, bsk) };
-        Ok(())
+        Ok(unsafe { self.discard_bootstrap_lwe_ciphertext_unchecked(output, input, acc, bsk) })
```
aPere3, line 132: It's weird because at some point clippy preferred the former way. Hopefully it does not seem to trigger here...

aPere3 (approved):
Good catch guys !

concrete-core/src/backends/cuda/implementation/engines/lwe_ciphertext_discarding_bootstrap.rs
agnesLeroy: Yes I remember that as well, but actually all the other engines are written in this format, so I thought I'd change those.

aPere3: Funny ! Anyway no problems on that front !

agnesLeroy:
Yes definitely!

IceTDrinker:
Otherwise @agnesLeroy for this one we can just disable the Clone part which is the really dangerous one, and keep the rest and follow up later ?

agnesLeroy:
Yes, sounds good.

PR #84, opened 2022-07-05
chore(rust): freeze nightly toolchain

IceTDrinker:
### Resolves:

closes https://github.com/zama-ai/concrete-core-internal/issues/192

### Description

- cargo task installs it automatically if not present
- all components available for this toolchain excpet rust analyzer preview

### Checklist 

* [ ] Check for breaking changes and add them to commit message following the conventional commit [specification][conventional-breaking]


IceTDrinker:
Tested on the M1 mac mini from the office and also Ben's M1 mac.

There are plans to get the mac mini as a runner for github actions to more easily iterate in CI (probably will run tests during the week-end as we only have a single M1 mac for now)

concrete-tasks/build.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        // For now we don't support native M1 compilation, so ask to use the x86_64 toolchain.
+        if is_arm64 {
+            Ok(format!("{}-x86_64-apple-darwin", toolchain_arg))
+        } else {
+            // Otherwise it's an x86 mac so just keep the default nightly toolchain
+            Ok(toolchain_arg.to_owned())
+        }
+    }
+}
```
aPere3, line 42: So here, If I understand correctly, on macos, it can be either:
+ `arm64` for m1 macs
+ `x86_64` for older macs

But whatever the case, we use `format!("{}-x86_64-apple-darwin", toolchain_arg)` is that right ? 

If so, maybe we can make this function simpler actually !

aPere3 (commented):
Thanks, for this PR ! 

On a general comment, I am not sure I am very fan of installing stuffs on behalf of the user. I was wondering if it would not be better to check if the toolchain is installed, and return a compile error with the command to be executed if not. 

Apart from this general comment. I have one small question on the code. 

Best,

Alex

concrete-tasks/build.rs
IceTDrinker: I guess so, by just formatting the x86 toolchain no matter what

IceTDrinker:
> Thanks, for this PR !
> 
> On a general comment, I am not sure I am very fan of installing stuffs on behalf of the user. I was wondering if it would not be better to check if the toolchain is installed, and return a compile error with the command to be executed if not.
> 
> Apart from this general comment. I have one small question on the code.
> 
> Best,
> 
> Alex

Yep, you are right, maybe I took a bit too liberally the fact that build.rs is an arbitrary code executor, I'll update the logic

IceTDrinker:
Ok it's a rustup 1.25 thing : https://github.com/rust-lang/rustup/issues/3031

aPere3:
Does that mean that there is no ways to make this thing work for now ?

IceTDrinker:
I have a partial workaround for 1.25, but it seems that other crates/build.rs are using wrappers around cargo so I'm seeing some build failures that look exactly like the one I showed you on Slack, so unless the fix gets propagated to all wrappers or rustup reverts the RUSTC thing I linked we currently have to require people to use rustup < 1.25

aPere3:
> I have a partial workaround for 1.25, but it seems that other crates/build.rs are using wrappers around cargo so I'm seeing some build failures that look exactly like the one I showed you on Slack, so unless the fix gets propagated to all wrappers or rustup reverts the RUSTC thing I linked we currently have to require people to use rustup < 1.25

:sob: That is very unfortunate ...

aPere3:
Oh yeah, you are right ! Rustup 1.25 just got out ! I missed it. Okay, then maybe we can use your workaround for now, and rework the solution when a new version of rustup shows up ?

IceTDrinker:
Yep, I'm on it as the main branch xtask is broken anyways for me, I'll make a quick and dirty solution and will revisit the topic later

IceTDrinker:
The last patch should allow rustup 1.25 users to use cargo xtasks, I'm really not sure why they suddenly decided that rustup was going to fill as many Env variables with hardcoded paths and libs linked to the rust toolchain 😐 (even saw some LD_PRELOAD for the toolchains, which seems a bit extreme, but I don't know what rustup used to do...)

IceTDrinker:
We'll need to work a bit on the CI to normalize things for toolchain setup (and the fact we need a stable and specific nightly toolchain) nothing major, but will make maintaining workflows easier

IceTDrinker:
simplified workaround after gathering infos on how the env was setup prior to rustup 1.25

Basically now rustc and rustdoc were selected through Env variables which was breaking wrappers, removing these env variables that are inserted by rustup 1.25 allows to get back the previous behavior

PR #85, opened 2022-07-06
fix(ci): fix workflow file format and concurrency for aws gpu tests

soonum:
### Description

Some under indented sections made that workflow impossible to be remotely triggered.

Also the concurrency group name has been enhanced to be able to run GPU and CPU in parallel. Now only a run with the git ref and the same AWS AMI would cancel any previous on-going run.

### Checklist 

~~* [ ] The PR description links to the related issue (to link an issue, use '#XXX'.)~~
~~* [ ] The tests on AWS have been launched and are successful (apply the `aws_test` to the PR to launch the tests on AWS)~~

### DIFF
--- a/.github/workflows/aws_tests_gpu_slab_beta.yml
+++ b/.github/workflows/aws_tests_gpu_slab_beta.yml
@@ -67,17 +67,16 @@ jobs:
       - name: Install Rust
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: nightly
           override: true
-      - name: Run concrete tests
-        run: cargo xtask test
+          components: rustfmt, clippy
+      - name: Clippy on cuda backend
+        run: |
+          cargo xtask check_clippy_cuda
 
-      - name: Test concrete-cuda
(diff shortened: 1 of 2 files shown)


agnesLeroy (commented):
Hi @soonum! Thanks a lot: does it mean that when triggering GPU tests we always have to specify the instance?

soonum:
No need to. In fact a default configuration is set in Slab config file, this is the same than what's used now with the label.
The concurrency naming change comes handy only when CPU and GPU are running at the same time AND if you want to launch two(or more) GPU tests suite but using different AMIs.

That said I should maybe add the EC2 product type to that naming since it's most likely this parameter one wants to change to test different hardware configuration.

soonum:
Wait before merging @agnesLeroy , I have to fix workflow steps before that.

agnesLeroy:
@soonum is it good now?

soonum:
It seems so, I just wait on Action results for GPU in my test PR.

agnesLeroy:
Let me know when it's ok, I'll approve and merge :)

agnesLeroy (approved):

PR #98, opened 2022-07-18
Feature/remove destroy

aPere3:
### Resolves: 

closes zama-ai/concrete-core-internal#312

### Checklist 


concrete-core-ffi/tests/test_lwe_bootstrap_w_bsk_serde.c
```diff
@@ -218,19 +218,19 @@ void bootstrap_view_buffers_test(void) {
...
-  default_engine_destroy_lwe_seeded_bootstrap_key_u64(default_engine, seeded_bsk);
-  default_engine_destroy_lwe_ciphertext_view_u64(default_engine, input_ct_as_view);
-  default_engine_destroy_lwe_ciphertext_mut_view_u64(default_engine, input_ct_as_mut_view);
-  default_engine_destroy_lwe_ciphertext_view_u64(default_engine, output_ct_as_view);
-  default_engine_destroy_lwe_ciphertext_mut_view_u64(default_engine, output_ct_as_mut_view);
-  default_engine_destroy_glwe_ciphertext_view_u64(default_engine, accumulator_as_view);
-  default_engine_destroy_glwe_ciphertext_mut_view_u64(default_engine, accumulator_as_mut_view);
-  fftw_engine_destroy_fftw_fourier_lwe_bootstrap_key_u64(fftw_engine, fbsk);
-  fftw_engine_destroy_fftw_fourier_lwe_bootstrap_key_u64(fftw_engine, deser_fbsk);
+  destroy_lwe_secret_key_u64( input_lwe_sk);
```
agnesLeroy, line 221: How come the engine name is not included in the name anymore?

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
```diff
@@ -56,3 +57,11 @@ impl<T: UnsignedInteger> CudaBootstrapKey<T> {
         self.d_ptr_vec[gpu_index.0 as usize]
     }
 }
+
```
agnesLeroy, line 60: Here I would add a comment stating that this implementation of Drop corresponds to the multi-GPU logic implemented in the Cuda backend, so one should be careful about it whenever wanting to change the multi-GPU logic. Same thing for the other Drop implementations.

agnesLeroy (commented):
Hi @aPere3! Thanks a lot for this PR, I'm so glad the destroys are going away :smile: I just have very minor comments/questions, otherwise it looks good to me. In the end the API will be much simpler like this! We'll update the user documentation at the end of the quarter I guess, it'll be easier to make one big pass over it before the release.

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
IceTDrinker: Is there maybe a way to tag pointers to know from which GPU they come so they are easier to drop ?

IMO the CUDA pointers should be wrapped in smart pointers (with GPU provenance e.g.) so that we don't have to do that manually and pointers clean themselves up without us having to do additional manual management, what do you think ?

concrete-core-ffi/src/backends/default/engines/destroy.rs
```diff
@@ -0,0 +1,270 @@
...
+/// This function is [checked](crate#safety-checked-and-unchecked-functions).
+#[no_mangle]
+pub unsafe extern "C" fn destroy_glwe_ciphertext_view_u64(
+    glwe_ciphertext_view: *mut GlweCiphertextView64,
+) -> c_int {
+    catch_panic(|| {
+        check_ptr_is_non_null_and_aligned(glwe_ciphertext_view).unwrap();
+
+        // Reconstruct the box, so that the memory is dropped at the end of the scope
+        let _glwe_ciphertext_view = Box::from_raw(glwe_ciphertext_view);
```
IceTDrinker, line 18: wondering about whether assigning the box here is useful, could maybe just be

`Box::from_raw(glwe_ciphertext_view);` ?

concrete-core-ffi/tests/test_lwe_bootstrap_w_bsk_serde.c
```diff
@@ -218,19 +218,19 @@ void bootstrap_view_buffers_test(void) {
...
+  destroy_lwe_secret_key_u64( output_lwe_sk);
```
IceTDrinker, line 222: could use a bit of clang format here 😄 

Not blocking, we don't have that in our makefile/xtasks

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
aPere3: That's what I wanted to do at first. Then, I thought that it would probably be simpler for gpu people to keep the gpu pointer type as it is, so that they can copy it at will. 

I think that attaching the gpu index to the pointer is better, but if we do it, we have to make sure that we do not implement `Clone` for the pointer type, otherwise we may have double free...

I'll let you decide @agnesLeroy !

IceTDrinker (commented):
My main comment would be to have a smart pointer for CUDA pointers so that we just have one implementation of Drop that gets re-used everywhere with GPU provenance so that it gets managed automatically

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
agnesLeroy: Yes and it would also facilitate future integration of a more flexible multi-GPU support.

IceTDrinker: hmm we can implement clone for the pointer type though and have it work properly, not sure I follow

aPere3: Okay, I'll do it then !

agnesLeroy: I don't know how to do that though :thinking:

IceTDrinker: do what ? 🙂

agnesLeroy: Sorry unsynchronized github comment

IceTDrinker:
Also this will need a rebase it seems 😬

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
aPere3: @IceTDrinker 

What I meant is that if we have that:
```rust
#[derive(Clone)] // or copy, even worse
pub struct CudaPointer{
    raw: *mut c_void
    gpu_inudex: usize
}

impl Drop for CudaPointer{
    fn drop(&mut self){
        cuda_free(self.raw, self.gpu_index)
    }
}
```
If we do a Clone, then drop one sibling, we free the memory, and the remaining pointer is dangling. We have to stick to the ownership model for that to work properly. I can see two ways to do it:
+ Statically : Do not implement `Clone` or `Copy`, this way the pointer must be moved, and ownership transferred
+ Dynamically: Have some kind of reference counting to only drop when the last reference disappears.

For obvious reasons, I am more in favor of option one :smile:

IceTDrinker: I would implement both Clone and Copy _manually_ as we are moving towards automatic memory management, wdyt ?

IceTDrinker: actually not copy which would be the door open to easy bugs

IceTDrinker: Clone would do a cudamalloc in my mind

aPere3: After discussing that, we chose to go for the smart pointer solution, without clone and copies.

concrete-core-ffi/tests/test_lwe_bootstrap_w_bsk_serde.c
aPere3: That is because the drop does not need the engine anymore. I removed the engine name from the function mangling.

concrete-core-ffi/src/backends/default/engines/destroy.rs
aPere3: You are right !

agnesLeroy:
@slab-ci cpu_test

agnesLeroy:
@slab-ci cpu_test

agnesLeroy:
@slab-ci cpu_test

agnesLeroy:
@slab-ci gpu_test

agnesLeroy (approved):

PR #109, opened 2022-07-21
test(tasks): check concrete-csprng behavior using statistics

soonum:
This new concrete task performs a behavior check of concrete-csprng
using the NIST statistical test suite tool. It assume NIST tool is
already installed on the machine. Several parameters can be tuned,
this include: random numbers generation duration, sequence length
and bitstreams value.

Also, clap crate version is now fixed to avoid frequent API breaking
changes when a new version is released.

### Checklist 

~~* [ ] The tests on AWS have been launched and are successful (apply the `aws_test` to the PR to launch the tests on AWS)~~


concrete-tasks/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 categories = ["do_not_publish"]
 
 [dependencies]
-clap = "3.1"
+clap = { version = "=3.2.12", features = ["derive"] }
```
IceTDrinker, line 8: Did you end up using the derive feature in the end ? I think you told me it did not mix well with the existing clap settings

concrete-tasks/src/main.rs
```diff
@@ -110,6 +111,42 @@ fn main() -> Result<(), std::io::Error> {
                 .about("Checks that clippy runs without warnings on the cuda backend"),
         )
         .subcommand(Command::new("check_fmt").about("Checks that rustfmt runs without warnings"))
+        .subcommand(
+            Command::new("check_csprng")
```
IceTDrinker, line 115: ould this command perhaps be generated in a function we would put in csprng_check.rs ?

concrete-tasks/src/main.rs
```diff
@@ -207,6 +244,9 @@ fn main() -> Result<(), std::io::Error> {
     if matches.subcommand_matches("check_fmt").is_some() {
         check::fmt()?;
     }
+    if matches.subcommand_matches("check_csprng").is_some() {
+        csprng_check::check(matches.subcommand_matches("check_csprng").unwrap())?;
+    }
```
IceTDrinker, line 249: maybe we want to add this function to tests @aPere3 wdyt ?

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+use log::{debug, info};
+use std::error::Error;
+use std::fs::File;
+use std::io::{self, BufRead, Error as IoError, ErrorKind, Write};
+use std::path::{Path, PathBuf};
+use std::process::{Child, Command, Stdio};
+use std::thread::sleep;
+use std::{fs, time};
+
+fn execute<T: Into<Stdio>>(
```
IceTDrinker, line 13: check utils.rs there is a function called cmd I think and a macro called cmd! that does more or less that

Unless you need something particular, which I'm not sure of yet

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+    Ok(parsed)
+}
+
+/// Check concrete-csprng behavior using a statistical approach provided by NIST test suite tool.
+pub fn check(args: &ArgMatches) -> Result<(), IoError> {
+    let generate_duration = parse_arg(args, "generate_duration")?;
+    let sequence_length = parse_arg(args, "sequence_length")?;
+    let bitstreams = parse_arg(args, "bitstreams")?;
+    let nist_tool_path = Path::new(args.value_of("nist_tool_dir").unwrap());
+    let build_path = Path::new("./csprng_check_build/");
```
IceTDrinker, line 186: this would need to be added to .gitignore if we are filling stuff in the repo root csprng_check_build

Btw there is an utils function to get the repo root, check utils.rs

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    let result = generate_random_numbers(build_path, nist_tool_path, generate_duration);
+    fs::remove_dir_all(build_path)
```
IceTDrinker, line 189: perhaps put that after the result check

Btw should we remove the build dir every time ?

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+///  concrete_path: path to directory containing concrete-core git repository
+/// duration: execution duration in seconds of concrete-csprng
+fn generate_random_numbers(
+    build_path: &Path,
+    nist_path: &Path,
+    duration: u64,
+) -> Result<(), Box<dyn Error>> {
+    fs::create_dir_all(build_path)?;
+    let build_cmd = format!(
+        "cargo {} {} -p concrete-csprng --bin generate --features=\"seeder_x86_64_rdseed generator_x86_64_aesni\" --target-dir {}",
```
IceTDrinker, line 59: `--features seeder_x86_64_rdseed --features generator_x86_64_aesni` works as well or you can use rust raw strings

https://rahul-thakoor.github.io/rust-raw-string-literals/

or even separate with commas IIRC `--features=seeder_x86_64_rdseed,generator_x86_64_aesni`

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+        "build", "--release", build_path.display()
```
IceTDrinker, line 60: this looks a bit odd for the command fomatting ?

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+    );
+    info!("Building concrete-csprng");
+    cmd!(<ENV_TARGET_NATIVE> &build_cmd)?;
+    info!("Successfully built concrete-csprng");
+
+    let sleep_duration = time::Duration::from_secs(duration);
+    // TODO Maybe chain join() calls to make path fully plateform independant?
```
IceTDrinker, line 67: should be fine, windows has been accepting forward slashes for a while and we don't even support windows for now

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+        &exec_path.display().to_string(),
+        None,
+        Some(&*ENV_TARGET_NATIVE),
+        Some(&*ROOT_DIR),
+        Stdio::inherit(),
+        Stdio::inherit(),
+        result_file.into(),
+    )?;
+    // Wait for enough data to be generated.
+    sleep(sleep_duration);
```
IceTDrinker, line 86: this is not precise in terms of the amount of data we'll generate

Maybe we need to update concrete-csprng/generate to take an argument saying how many bytes we want to generate

wdyt ?

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+    proc.kill()?;
```
IceTDrinker, line 87: can we sigterm rather than kill ?

concrete-tasks/src/main.rs
```diff
@@ -110,6 +111,42 @@ fn main() -> Result<(), std::io::Error> {
...
+        .subcommand(
+            Command::new("check_csprng")
+                .about("Checks concrete-csprng behavior with statistical approach")
+                .arg(
+                    Arg::new("nist_tool_dir")
+                        .short('n')
+                        .long("nist_tool_dir")
+                        .takes_value(true)
+                        .required(true)
+                        .help("Path to NIST statistical tool directory"),
```
IceTDrinker, line 123: we don't install it ourselves with a shell script or task to always find it in the same place ?

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,207 @@
...
+        Stdio::piped(),
+    )?;
+    let mut stdin = proc.stdin.take().expect("Failed to get stdin");
+    let half_a_sec = time::Duration::from_millis(500);
+    for value in inputs {
+        stdin.write_all(format!("{}\n", value).as_bytes())?;
+        stdin.flush()?;
+        // Give some time for the input to be processed.
+        sleep(half_a_sec);
+    }
```
IceTDrinker, line 134: I think you want stdio inherit

IceTDrinker (changes requested):
Hello @soonum thanks a lot for the work on this !

I have a few comments (mostly related I think to the tools we already have in tasks).

I think we are missing a script/task to install the nist tool automatically

There are a few things code-wise to modify before getting onto the stage 2 of this review 🙂

soonum:
> I think we are missing a script/task to install the nist tool automatically

In fact I was thinking about doing another commit purely CI related once this task is merged. See the [associated issue comment](https://github.com/zama-ai/concrete-core-internal/issues/247#issuecomment-1191105809)

concrete-tasks/src/csprng_check.rs
soonum: I admit that SIGKILL is a bit harsh but there is no SIGTERM or signal sending method is `std::process::Child` for now. :cry:

IceTDrinker:
> > I think we are missing a script/task to install the nist tool automatically
> 
> In fact I was thinking about doing another commit purely CI related once this task is merged. See the [associated issue comment](https://github.com/zama-ai/concrete-core-internal/issues/247#issuecomment-1191105809)

Then I would do them the other way around or have two commits in this PR ? The reasons being we will be building on top of the tool to install the NIST tool ? 🙂

concrete-tasks/src/csprng_check.rs
soonum: That would be great to specify a number of bytes instead of a duration, that way I don't need to kill the child and thus I would be able to use `cmd!` macro instead of my `execute()` implementation.
Is this change in concrete-csprng doable easily?

concrete-tasks/src/csprng_check.rs
IceTDrinker: ah well then kill if with 🔥  😄

concrete-tasks/src/csprng_check.rs
IceTDrinker: I think so yeah, it's in the generate binary, don't remember if it takes arguments today

concrete-tasks/Cargo.toml
soonum: You're right, i forgot to remove the feature.

concrete-tasks/src/main.rs
soonum: That's a great idea, it would keep command declaration in main.rs more concise.

concrete-tasks/src/csprng_check.rs
soonum: I want to be able to manipulate standard streams, I could add these arguments in utils.rs if that's okay.

IceTDrinker: just to understand what do you need to manipulate in the streams ?

concrete-tasks/src/csprng_check.rs
soonum: I remove the directory and its content right after the function call to avoid dealing with errors from next function calls.
That said, if we don't remove the directory then the problem is solved. 
Removing directory is handy for people running this task on their local machine. In CI, we don't care about removing since the instance/VM would be destroyed at the end of the test suite.

IceTDrinker: might be easier to see with the build script for nist :)

concrete-tasks/src/csprng_check.rs
soonum: You mean adding the custom build path .gitignore right ?

concrete-tasks/src/csprng_check.rs
soonum: Thanks for the tips :+1: 
Something came into my mind, what about the behavior of this command once the M1 support would be available? Those features won't work right ?

concrete-tasks/src/csprng_check.rs
IceTDrinker: yes, looks like it will be in the repo here ?

concrete-tasks/src/csprng_check.rs
IceTDrinker: correct, but as this is not merged yet, I don't think you should worry too much about it currently 🙂

concrete-tasks/src/csprng_check.rs
soonum: Yeah it's a leftover from when I used cargo to run the program after buidling it. What do you think of:
```rust
let build_cmd = format!(
    "cargo build --release -p concrete-csprng --bin generate --features=\"seeder_x86_64_rdseed generator_x86_64_aesni\" --target-dir {}",
    build_path.display()
);
```

concrete-tasks/src/csprng_check.rs
soonum: First, I need to redirect the output of concrete-csprng into a file. Second, NIST tool is interactive so I have to manipulate STDIN.

concrete-tasks/src/csprng_check.rs
soonum: Indeed.

concrete-tasks/src/csprng_check.rs
IceTDrinker: understood, makes sense

then I would maybe put that in utils.rs indicating this allows to interact with interactive programs from rust

concrete-tasks/src/csprng_check.rs
soonum: From a separation of concerns standpoint, NIST build script has nothing to do with that, we are building concrete-csprng in this directory and that's all. This has nothing to do with NIST.
Or am I misunderstanding the issue here?

concrete-tasks/src/csprng_check.rs
soonum: I like this kind of reply :smirk:

concrete-tasks/src/csprng_check.rs
IceTDrinker: What I mean is it's hard to see what the problems might be with removing the build dir without the build script :)

concrete-tasks/src/csprng_check.rs
IceTDrinker: yep looks better this way

concrete-tasks/src/csprng_check.rs
soonum: I belive, if I inherit STDIN, that the xtask will prompt for user input. We don't want that since we want to automate the process.

IceTDrinker: Yes you are right, the other discussion answered that question 👍

concrete-tasks/src/csprng_check.rs
soonum: Oh, I get it. This directory has a unique purpose, containing the build artifacts from concrete-csprng. NIST tool should be built in there I think.

IceTDrinker: hmm, but the generate binary would be built in our target/release dir right ?

concrete-tasks/src/main.rs
soonum: The NIST build script should be not entangled with this task, since it has "nothing" to do with concrete-core repo directly, it's just a CI dependency. That's why I choose to use a command argument to locate NIST tool.
But I'm open to other visions.

IceTDrinker: I think it should be at least easily installable through a make command e.g. as people will need it on their machines to test the csprng e.g.

basically anything in the CI that might be used by a dev needs to be easily installable by a dev on his machine, either automagically or through a command they can run and be done with it IMO

soonum:
Guys, I've pushed the modifications for concrete-csprng to be able to generate a fixed number of bytes. This commit is intended to be the first to be merged once this review is done. I wanted to separate it from the rest since it adds a feature to the concrete-csprng crate.

concrete-csprng/Cargo.toml
```diff
@@ -13,6 +13,7 @@ keywords = ["fully", "homomorphic", "encryption", "fhe", "cryptography"]
 
 [dependencies]
 aes-soft = "0.6.4"
+clap = "=3.2.12"
```
IceTDrinker, line 16: is it possible to add the dependency only for the binary ?

IceTDrinker:
For the binary update the code looks ok and can be put in a separate PR potentially (not sure how you want to handle it) but I have that one question about putting clap as dependency of the generate binary only

concrete-csprng/Cargo.toml
soonum: what do you mean? Do I have to put it in `[[bin]]` section and then add some configuration directives to avoid building with clap in during dev ?

soonum:
Hmmm since there is a new feature in concrete-csprng, shouldn't we bump the version of the crate in the meantime ? Something like `0.3.0` ?

IceTDrinker:
> Hmmm since there is a new feature in concrete-csprng, shouldn't we bump the version of the crate in the meantime ? Something like `0.3.0` ?

it's not a breaking change for the lib, so at worst I'd say a 0.2.x release

concrete-csprng/Cargo.toml
IceTDrinker: Let me check, I'm not sure about how cargo handles that

IceTDrinker: https://stackoverflow.com/a/46017284/18088947

Seems there is no direct way of defining deps for binaries and libs in the same crate

soonum: I see, what should we do ? Add a workaround or then left the dependency as is ?

concrete-tasks/src/main.rs
soonum: Done in [a792a48](https://github.com/zama-ai/concrete-core/pull/109/commits/a792a48515431dd9320dd9953f8614cda354f162)

concrete-tasks/src/csprng_check.rs
soonum: Now th NIST tool is built in `csprng_check_builds/` along with concrete-csprng.

soonum:
> it's not a breaking change for the lib nor for the binary, so at worst I'd say a 0.2.x release

Changed for `0.2.1`, it will visible at the next push.

Makefile
```diff
@@ -16,6 +16,16 @@ check_tasks_rust_toolchain:
 		exit 1; \
 	fi
 
+BUILD_DIR := $(CURDIR)/csprng_check_builds
```
IceTDrinker, line 19: CSPRNG_BUILD_DIR
or even
CSPRNG_CHECK_BUILD_DIR

Makefile
```diff
@@ -16,6 +16,16 @@ check_tasks_rust_toolchain:
...
+NIST_DIR := $(BUILD_DIR)/nist_sts
+
+.PHONY: install_nist_test_tool # Install NIST statistical test suite tool used to check concrete-csprng behavior
+install_nist_test_tool:
+	mkdir -p $(NIST_DIR)
```
IceTDrinker, line 24: double quote around all $() perhaps ? I know I'm annoying with that thing 😅

Makefile
```diff
@@ -16,6 +16,16 @@ check_tasks_rust_toolchain:
...
+	cd $(BUILD_DIR) && wget "https://csrc.nist.gov/CSRC/media/Projects/Random-Bit-Generation/documents/sts-2_1_2.zip" && unzip -q sts-2_1_2.zip
+	mv $(BUILD_DIR)/sts-2.1.2/sts-2.1.2 $(NIST_DIR)
```
IceTDrinker, line 26: can we have a sha256 check here please ?

Makefile
```diff
@@ -16,6 +16,16 @@ check_tasks_rust_toolchain:
 		exit 1; \
 	fi
 
+BUILD_DIR := $(CURDIR)/csprng_check_builds
+NIST_DIR := $(BUILD_DIR)/nist_sts
```
IceTDrinker, line 20: this is a build dir as well as we call make

.gitignore
```diff
@@ -26,3 +26,6 @@ Cargo.lock
 
 # concrete-core-ffi build artifacts
 /concrete-core-ffi/build/
+
+# concrete-tasks check_csprng artifacts directory
+/csprng_check_builds
```
IceTDrinker, line 31: I'm wondering if we might be able to put ourselves in target ? do you know what the best practices are around the target dir in rust for custom/additional builds ?

concrete-tasks/src/csprng_check.rs
```diff
@@ -1,91 +1,82 @@
...
     let results_path = nist_path.join("data/csprng_results");
     let result_file = File::create(results_path)?;
-    // The binary output from `cargo build` is directly used since `cargo run` would spawn
-    // the program as a child process. Thus doing Child.kill() would only kill `cargo run` command
-    // but not the underlying program that is being executed. As a consequence, concrete-csprng
-    // would run indefinitely.
-    let exec_path = build_path.join("release/generate");
+    let run_cmd = format!(
+        "{} -b {}",
+        build_path.join("release/generate").display(),
```
IceTDrinker, line 74: I understand the .display as paths may not be valid utf8, is there a way to call the command without doing that display thing ? 🤔 

I'm guessing probably not but wanted to ask anyways :)

concrete-tasks/src/csprng_check.rs
```diff
@@ -132,15 +119,13 @@ fn run_test_suite(
         // Give some time for the input to be processed.
         sleep(half_a_sec);
```
IceTDrinker, line 120: Is sleeping here mandatory ?

I know you told me we interact with the tool in interactive mode but I feel like with the \n after the values we should be fine ?

Also there really is no way to launch the tool in cli mode/uninteractive ? 🤔 (sorry I really don't know that NIST tool thing)

concrete-tasks/src/main.rs
```diff
@@ -36,11 +36,44 @@ lazy_static! {
 
 #[macro_export]
 macro_rules! cmd {
+    (<$env: ident> $cmd: expr, $cwd: expr, $stdin: expr, $stderr: expr, $stdout: expr, $return_handle: expr) => {
```
IceTDrinker, line 39: why is this change needed ? I'm a bit confused

concrete-tasks/src/utils.rs
```diff
@@ -2,25 +2,34 @@ use log::{debug, info};
...
     }
     let mut command = Command::new("sh");
     command
         .arg("-c")
         .arg(cmd)
-        .stderr(Stdio::inherit())
-        .stdout(Stdio::inherit());
+        .stdin(stdin.unwrap_or(Stdio::piped()))
+        .stderr(stderr.unwrap_or(Stdio::piped()))
+        .stdout(stdout.unwrap_or(Stdio::piped()));
```
IceTDrinker, line 32: We should keep the default behavior here I would say

IceTDrinker (commented):
Thanks a lot @soonum looking much better 🙂 

Could you add the csprng check to the CPU workflow so that we can check it works as expected ?

I think I only visualized the last diff, so not sure if it was already done

concrete-tasks/src/csprng_check.rs
soonum: I used the `to_str()` which returns an `Option<str>`. I've made some assumptions, like the `build_path` is guaranteed to contains only valid utf8 chars (since its value is crafted by us in the code). Conversely, I explicitly check for correctness regarding `nist_tool_path`.

concrete-tasks/src/main.rs
soonum: I added this flavor of the macro to be able to modify standard streams without changing all the `cmd!` across the whole crate.

concrete-tasks/src/csprng_check.rs
soonum: Sadly we can't avoid having the interactive mode when running the NIST tool.
I put this sleep in place because I was having some troubles back when the script was written in Python. Anyway I tested without the sleep and it works like a charm. :tada:

IceTDrinker: Great let’s keep like this then!

concrete-tasks/src/csprng_check.rs
IceTDrinker: Sounds good

concrete-tasks/src/main.rs
IceTDrinker: Ok cool

Makefile
soonum: I'm not sure that is necessary in this case, definition of the variable is given just above the recipe so we can check that there is no bad intentions when performing the interpolation.

soonum:
> Could you add the csprng check to the CPU workflow so that we can check it works as expected ?

What do you mean by "CPU workflow" ? build.yaml ? aws_tests.yml ? Both ?

IceTDrinker:
> > Could you add the csprng check to the CPU workflow so that we can check it works as expected ?
> 
> What do you mean by "CPU workflow" ? build.yaml ? aws_tests.yml ? Both ?

I think we want it in cargo xtask test, is it possible to also save a copy of the NIST archive somewhere in case it was to mysteriously disappear ?

.github/workflows/build.yml
```diff
@@ -24,6 +24,15 @@ jobs:
...
         if: ${{ !cancelled() }}
         run: cargo xtask build_simd_crates
+      - name: Install NIST statistical test suite
+        if: ${{ !cancelled() }}
+        run: |
+          make install_nist_test_tool
+      - name: Check concrete-csprng
+        if: ${{ !cancelled() }}
+        run: |
+          cargo xtask check_csprng --nist_tool_dir ${{ GITHUB_WORKSPACE }}/csprng_check_builds/nist_sts/sts-2.1.2 --generated_bytes <FIXME: put here a meaningful value of bytes to generate>
```
IceTDrinker, line 34: @blowfish880 do you have an opinion on how many bytes we should generate with the CSPRNG to check it with the NIST tool suite ?

A reasonnable amount if possible to avoid spending too much time in that test but still enough to be meaningful

@soonum to run the tests in our ci : 10 000 000 is a good starting position as I think it's what we use in some of our tests currently

Makefile
```diff
@@ -16,6 +16,21 @@ check_tasks_rust_toolchain:
...
+CSPRNG_CHECK_BUILD_DIR := $(CURDIR)/csprng_check_builds
+NIST_BUILD_DIR := $(CSPRNG_CHECK_BUILD_DIR)/nist_sts
+
+.PHONY: install_nist_test_tool # Install NIST statistical test suite tool used to check concrete-csprng behavior
+install_nist_test_tool:
+	mkdir -p $(NIST_BUILD_DIR)
+	cd $(CSPRNG_CHECK_BUILD_DIR); \
+	wget "https://csrc.nist.gov/CSRC/media/Projects/Random-Bit-Generation/documents/sts-2_1_2.zip" && \
+	echo "0238d2f1d26e120e3cc748ed2d4c674cdc636de37fc4027c76cc2a394fff9157  sts-2_1_2.zip" >> checksum && \
+	sha256sum -c checksum && \
```
IceTDrinker, line 28: this won't be compatible with macOS for the sha256sum I think

sha256sum is equivalent to `shasum -a 256` not sure for the -c flag

concrete-csprng/Cargo.toml
IceTDrinker: making it optional with a feature like "csprng_generate_bin_clap" at least so it does not get pulled with CSPRNG maybe ? :/

concrete-csprng/src/main.rs
```diff
@@ -1,20 +1,50 @@
...
     let mut stdout = stdout();
     let mut buffer = [0u8; 16];
-    loop {
-        buffer
-            .iter_mut()
-            .zip(&mut generator)
-            .for_each(|(b, g)| *b = g);
-        stdout.write_all(&buffer).unwrap();
-    }
+    match matches.value_of("bytes_total") {
```
IceTDrinker, line 30: could we have two function for the two cases ? 🙂

concrete-tasks/src/csprng_check.rs
```diff
@@ -0,0 +1,195 @@
...
+                .takes_value(true)
+                .required(true)
+                .help("Path to NIST statistical tool directory"),
+        )
+        .arg(
+            Arg::new("generated_bytes")
+                .short('g')
+                .long("generated_bytes")
+                .takes_value(true)
+                .default_value("4096") // FIXME replace by a realistic but large default value
```
IceTDrinker, line 27: maybe 10 000 000

IceTDrinker (commented):
This looks much better 🙂 

Main thing now is having an actual test in CI, I have a proposal for the number of bytes for that.

If you want me to test it locally let me know !

concrete-tasks/src/csprng_check.rs
soonum: 80.000.000 bytes are generated with 10.000.000 of `f64`.

.github/workflows/build.yml
soonum: I'm waiting for @blowfish880 advice before putting any value at this place.
FYI, I've tried 10.000.000 and 80.000.000, both were generated in less than one second on my laptop and both failed at "universal" statistical test.

concrete-csprng/src/main.rs
```diff
@@ -12,6 +14,38 @@ fn write_bytes(buffer: &mut [u8], generator: &mut AesniRandomGenerator, stdout:
...
+    buffer: &mut [u8],
+    generator: &mut AesniRandomGenerator,
+    stdout: &mut Stdout,
+) {
+    loop {
+        write_bytes(buffer, generator, stdout);
+    }
+}
+
+#[cfg(feature = "csprng_generate_bin_clap")]
```
IceTDrinker, line 27: is it possible to require the feature for a binary ? I think I saw something like that somewhere in a Cargo.toml, would allow to remove those cfg clauses

IceTDrinker:
@soonum what were we waiting for here ? I don't recall 🤔

soonum:
We are waiting for recommendations from @blowfish880 for the number of bytes to generate to pass the `universal` test case.
But if it keeps failing we could eventually silence this test case if it doesn't have a great impact on qualifying the statistical correctness of our random number generator.

.github/workflows/build.yml
blowfish880: Sorry, that slipped past my attention. I'm looking into it.

blowfish880: According to the documentation, it seems that `sequence_length = 10^6` and `bitstreams=200` are sensible parameters. Accordingly, we need to generate `sequence_length*bitstreams = 2*10^8` many bits, which corresponds to 25 MiB of random data.

soonum: With a `sequence_length = 400 000` and `bitstreams = 200` the test takes around 7 minutes on my laptop. With your values @blowfish880 the test duration is around 20 minutes.

The question is, is waiting 7 more minutes in the build action is OK ? Or do we want to skip the `universal` test case ?
In case of the latter, we could reduce the numbers used as inputs and thus having an execution time below 2'.

IceTDrinker: 20 minutes for 25 MiB of random data ? that seems odd, is it the csprng or the test suite that is slow ?

IceTDrinker: does it work for the universal test btw ?

soonum: Yes it is the time it takes to NIST to perform all its checks. The `universal` test case passes in both cases (7' and 20')

IceTDrinker: no chance to maybe use multithreading on that nist tool ?

soonum: Nope sadly.

IceTDrinker: Otherwise we run this in an other github runner in parallel

soonum: This could work too.

soonum: @agnesLeroy what do you think of the solution of using another action to run csprng checks ?

IceTDrinker: can be just a self hosted one, as long as it takes less than 1 hour we won't see it in our test times

agnesLeroy: @IceTDrinker @soonum if it's 20 minutes having it on a separate runner makes sense. Actually having an action we only trigger every week is enough, we don't need to trigger this check on every PR.

soonum: Ok I'll create a new workflow file for csprng checks then. I'll schedule it twice a week for now, let's say during Sunday night and Wednesday night. How does it sound to you ?

However, I'm not sure I understand your second idea. When you say:

>  we could also keep only the doc test build on every PR but run them only nightly or weekly.

Do you mean that you want to remove the doc builds performed on each push in a PR and instead do it on the dev branch every once in a while ?

soonum: I've run tests on my laptop with different parameters for `sequence_length` (`bitstreams = 200` each time), here are the results:

- `sequence_length = 10⁶` -> 2h29'  | Failing tests : NonOverLappingTemplate
- `sequence_length = 10⁵` -> 14'      | Failing tests : NonOverLappingTemplate
- `sequence_length = 5*10⁵` -> 8'    | Failing tests : CumulativeSum
- `sequence_length = 4*10⁵` -> 6'    | Failing tests : None

Do these results reveal something about `concrete-csprng` ?
In any case, should we use the last parameter choice which is fast and doesn't return any error?

agnesLeroy: @soonum "however, I'm not sure I understand your second idea. When you say:

    we could also keep only the doc test build on every PR but run them only nightly or weekly.

Do you mean that you want to remove the doc builds performed on each push in a PR and instead do it on the dev branch every once in a while ?"

Currently for each AWS run we execute `cargo test --doc -p concrete-core --all-features`. What I was trying to say is that we could move that task out of the usual jobs and into a bi-weekly check. The reason is that checking doc tests compile should actually be enough for us on a daily basis. They don't really check anything useful so running them (which takes time) could be done on a less frequent basis, to reduce tests time in our daily workflow.

IceTDrinker: > I've run tests on my laptop with different parameters for `sequence_length` (`bitstreams = 200` each time), here are the results:
> 
> * `sequence_length = 10⁶` -> 14'      | Failing tests : NonOverLappingTemplate
> * `sequence_length = 5*10⁵` -> 8'    | Failing tests : CumulativeSum
> * `sequence_length = 4*10⁵` -> 6'    | Failing tests : None
> 
> Do these results reveal something about `concrete-csprng` ? In any case, should we use the last parameter choice which is fast and doesn't return any error?

Probably, but as we use a construct that has been validated as a CSPRNG I'm not sure what to think about these failing tests 🤔 

Btw do repeated tests yield the same results ?

soonum: @agnesLeroy I don't find this doc test in this branch. The only occurrence I have is: 
`cargo test --doc -p concrete-core --features=backend_cuda -- backends::cuda`

Is this the one you mean ?

agnesLeroy: @soonum actually for Cuda we need to execute the doc tests so I'd leave it :sweat_smile: and @IceTDrinker realized our serialization tests are in the doc tests so you can forget about what I said and leave the doc tests' execution as it is for now :sweat_smile: sorry about this!

IceTDrinker: cargo test should run both classic and doc tests

concrete-tasks/src/main.rs
IceTDrinker: we will most likely run it in a github runner in parallel to tests to avoid slowing things down

soonum: So finally I have to add a new workflow file right ?

soonum:
I don't fully understand your question.

> about this PR, do several tests all yield same the same results from NIST @soonum ?

Do you mean that several workflows execute the csprng checks ?

concrete-tasks/src/main.rs
IceTDrinker: not necessarily a workflow file, but it needs a job somewhere to run a separate runner

IceTDrinker:
No I meant, does running the same test over and over again give consistent result, i.e. all pass for example ?

soonum:
> No I meant, does running the same test over and over again give consistent result, i.e. all pass for example ?

Nope it doesn't. I've run a small batch of tests, with the same parameters, 7 tests out of ten fails.

IceTDrinker:
> > No I meant, does running the same test over and over again give consistent result, i.e. all pass for example ?
> 
> Nope it doesn't. I've run a small batch of tests, with the same parameters, 7 tests out of ten fails.

yikes... so we can't merge it for now IMO :/

soonum:
@blowfish880 is currently looking into it to know how we can have repeatable results.

PR #130, opened 2022-08-19
Feature/circuit bootstrapping

IceTDrinker:
### Resolves:

closes https://github.com/zama-ai/concrete-core-internal/issues/352

### Description

Add private implementation for circuit bootstrap, slight rework of the FPKSK to take a reference to a function/closure to avoid the issue of the closure size while still being able to pass a function (less tedious than writing polynomials by hand).

Slight rework of the extract bits to be able to pass an output container directly

### Checklist 

### DIFF
--- a/concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
+++ b/concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
@@ -43,7 +43,7 @@ pub struct LweCiphertextVectorGlweCiphertextDiscardingPrivateFunctionalPackingKe
     pub output_polynomial_size: PolynomialSize,
     pub decomposition_level: DecompositionLevelCount,
     pub decomposition_base_log: DecompositionBaseLog,
-    pub function_lipschitz_bound: usize,
+    pub function_log_scalar: usize,
 }
 
 impl<
@@ -142,7 +142,7 @@ where
                     output_polynomial_size: PolynomialSize(256),
                     decomposition_level: DecompositionLevelCount(3),
                     decomposition_base_log: DecompositionBaseLog(7),
-                    function_lipschitz_bound: 1,
+                    function_log_scalar: 0,
                 },
                 LweCiphertextVectorGlweCiphertextDiscardingPrivateFunctionalPackingKeyswitchParameters {
                     input_lwe_noise: Variance(
@@ -157,7 +157,7 @@ where
                     output_polynomial_size: PolynomialSize(256),
                     decomposition_level: DecompositionLevelCount(3),
                     decomposition_base_log: DecompositionBaseLog(7),
-                    function_lipschitz_bound: 100,
+                    function_log_scalar: 4,
                 },
             ]
             .into_iter(),
@@ -181,18 +181,19 @@ where
             parameters.output_glwe_dimension,
             parameters.output_polynomial_size,
         );
-        let raw_cleartext_vector = Precision::Raw::uniform_zero_centered_vec(
-            parameters.function_lipschitz_bound,
-            parameters.output_polynomial_size.0,
-        );
+        let raw_cleartext_vector =
+            Precision::Raw::uniform_zero_centered_vec(1, parameters.output_polynomial_size.0);
         let proto_cleartext_vector =
             maker.transform_raw_vec_to_cleartext_vector(&raw_cleartext_vector);
+        let scalar = Precision::Raw::power_of_two(parameters.function_log_scalar);
+
         let proto_pfpksk = maker.new_private_functional_packing_keyswitch_key(
             &proto_secret_key_input,
             &proto_secret_key_output,
             parameters.decomposition_level,
             parameters.decomposition_base_log,
             parameters.pfpksk_noise,
+            &|x| scalar * x,
             &proto_cleartext_vector,
         );
         (
@@ -319,6 +320,7 @@ where
         _maker: &mut Maker,
         _repetition_proto: &Self::RepetitionPrototypes,
     ) -> Self::Criteria {
+        let polynomial_infinity_norm = 1.;
         let predicted_variance =
             fix_estimate_private_functional_keyswitch_noise_lwe_to_glwe_with_constant_terms::<
                 Precision::Raw,
@@ -331,7 +333,7 @@ where
                 parameters.pfpksk_noise,
                 parameters.decomposition_base_log,
                 parameters.decomposition_level,
-                parameters.function_lipschitz_bound as f64,
+                2_f64.powi(parameters.function_log_scalar as i32) * polynomial_infinity_norm,
             );
         (Variance(
             predicted_variance.0 * parameters.input_lwe_count.0 as f64,
--- a/concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
+++ b/concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
@@ -1,13 +1,16 @@
 //! Primitives for the so-called Wop-PBS (Without Padding Programmable Bootstrapping)
 
 use crate::backends::fftw::private::crypto::bootstrap::{FourierBootstrapKey, FourierBuffers};
-use crate::backends::fftw::private::math::fft::{AlignedVec, Complex64};
+use crate::backends::fftw::private::math::fft::Complex64;
 use crate::commons::crypto::encoding::Cleartext;
-use crate::commons::crypto::glwe::GlweCiphertext;
+use crate::commons::crypto::ggsw::StandardGgswCiphertext;
+use crate::commons::crypto::glwe::{GlweCiphertext, PrivateFunctionalPackingKeyswitchKeyList};
 use crate::commons::crypto::lwe::{LweCiphertext, LweKeyswitchKey, LweList};
-use crate::commons::math::tensor::{AsMutTensor, AsRefTensor};
+use crate::commons::math::tensor::{AsMutTensor, AsRefSlice, AsRefTensor};
 use crate::commons::math::torus::UnsignedTorus;
-use concrete_commons::parameters::{CiphertextCount, DeltaLog, ExtractedBitsCount, LweDimension};
+use concrete_commons::parameters::{
+    DecompositionBaseLog, DecompositionLevelCount, DeltaLog, ExtractedBitsCount, LweDimension,
+};
 
 #[cfg(test)]
 mod test;
@@ -18,16 +21,17 @@ mod test;
 /// Ouput bits are ordered from the MSB to the LSB. Each one of them is output in a distinct LWE
 /// ciphertext, containing the encryption of the bit scaled by q/2 (i.e., the most significant bit
 /// in the plaintext representation).
-pub fn extract_bits<Scalar>(
+pub fn extract_bits<Scalar, C1>(
     delta_log: DeltaLog,
+    lwe_list_out: &mut LweList<Vec<Scalar>>,
     lwe_in: &LweCiphertext<Vec<Scalar>>,
     ksk: &LweKeyswitchKey<Vec<Scalar>>,
-    fourier_bsk: &FourierBootstrapKey<AlignedVec<Complex64>, Scalar>,
+    fourier_bsk: &FourierBootstrapKey<C1, Scalar>,
     buffers: &mut FourierBuffers<Scalar>,
     number_of_bits_to_extract: ExtractedBitsCount,
-) -> LweList<Vec<Scalar>>
-where
+) where
     Scalar: UnsignedTorus,
+    FourierBootstrapKey<C1, Scalar>: AsRefTensor<Element = Complex64>,
 {
     let ciphertext_n_bits = Scalar::BITS;
     let number_of_bits_to_extract = number_of_bits_to_extract.0;
@@ -40,6 +44,18 @@ where
         ciphertext_n_bits,
         delta_log.0,
         ciphertext_n_bits - delta_log.0,
+    );
+    debug_assert!(
+        lwe_list_out.lwe_size() == ksk.lwe_size(),
+        "lwe_list_out needs to have an lwe_size of {}, got {}",
+        ksk.lwe_size().0,
+        lwe_list_out.lwe_size().0,
+    );
+    debug_assert!(
+        lwe_list_out.count().0 == number_of_bits_to_extract,
+        "lwe_list_out needs to have a ciphertext count of {}, got {}",
+        number_of_bits_to_extract,
+        lwe_list_out.count().0,
     );
 
     let polynomial_size = fourier_bsk.polynomial_size();
@@ -57,15 +73,8 @@ where
         LweDimension(glwe_dimension.0 * polynomial_size.0).to_lwe_size(),
     );
 
-    // Output List
-    let mut output_lwe_list = LweList::allocate(
-        Scalar::ZERO,
-        ksk.lwe_size(),
-        CiphertextCount(number_of_bits_to_extract),
-    );
-
     // We iterate on the list in reverse as we want to store the extracted MSB at index 0
-    for (bit_idx, mut output_ct) in output_lwe_list.ciphertext_iter_mut().rev().enumerate() {
+    for (bit_idx, mut output_ct) in lwe_list_out.ciphertext_iter_mut().rev().enumerate() {
         // Shift on padding bit
         lwe_bit_left_shift_buffer.fill_with_scalar_mul(
             &lwe_in_buffer,
@@ -121,6 +130,169 @@ where
         // Remove the extracted bit from the initial LWE to get a 0 at the extracted bit location.
         lwe_in_buffer.update_with_sub(&lwe_out_pbs_buffer);
     }
+}
 
-    output_lwe_list
+/// Circuit bootstrapping for binary messages, i.e. containing only one bit of message
+///
+/// The output GGSW ciphertext `ggsw_out` decomposition base log and level count are used as the
+/// circuit_bootstrap_binary decomposition base log and level count.
+pub fn circuit_bootstrap_binary<Scalar, C1, C2, C3, C4>(
+    fourier_bsk: &FourierBootstrapKey<C1, Scalar>,
+    lwe_in: &LweCiphertext<C2>,
+    ggsw_out: &mut StandardGgswCiphertext<C3>,
+    buffers: &mut FourierBuffers<Scalar>,
+    delta_log: DeltaLog,
+    fpksk_list: &PrivateFunctionalPackingKeyswitchKeyList<C4>,
+) where
+    Scalar: UnsignedTorus,
+    FourierBootstrapKey<C1, Scalar>: AsRefTensor<Element = Complex64>,
+    C2: AsRefSlice<Element = Scalar>,
+    StandardGgswCiphertext<C3>: AsMutTensor<Element = Scalar>,
+    PrivateFunctionalPackingKeyswitchKeyList<C4>: AsRefTensor<Element = Scalar>,
+{
+    let level_cbs = ggsw_out.decomposition_level_count();
+    let base_log_cbs = ggsw_out.decomposition_base_log();
+
+    debug_assert!(
+        level_cbs.0 >= 1,
+        "level_cbs needs to be >= 1, got {}",
+        level_cbs.0
+    );
+    debug_assert!(
+        base_log_cbs.0 >= 1,
+        "base_log_cbs needs to be >= 1, got {}",
+        base_log_cbs.0
+    );
+
+    let bsk_glwe_dimension = fourier_bsk.glwe_size().to_glwe_dimension();
+    let bsk_polynomial_size = fourier_bsk.polynomial_size();
+
+    let fpksk_input_lwe_key_dimension = fpksk_list.input_lwe_key_dimension();
+
+    debug_assert!(
+        fpksk_input_lwe_key_dimension.0 == bsk_polynomial_size.0 * bsk_glwe_dimension.0,
+        "The fourier_bsk polynomial_size, got {}, must be equal to the fpksk \
+        input_lwe_key_dimension, got {}",
+        bsk_polynomial_size.0 * bsk_glwe_dimension.0,
+        fpksk_input_lwe_key_dimension.0
+    );
+
+    let fpksk_output_polynomial_size = fpksk_list.output_polynomial_size();
+    let fpksk_output_glwe_key_dimension = fpksk_list.output_glwe_key_dimension();
+
+    debug_assert!(
+        ggsw_out.polynomial_size() == fpksk_output_polynomial_size,
+        "The output GGSW ciphertext needs to have the same polynomial size as the fpksks, \
+        got {}, expeceted {}",
+        ggsw_out.polynomial_size().0,
+        fpksk_output_polynomial_size.0
+    );
+
+    debug_assert!(
+        ggsw_out.glwe_size().to_glwe_dimension() == fpksk_output_glwe_key_dimension,
+        "The output GGSW ciphertext needs to have the same GLWE dimension as the fpksks, \
+        got {}, expeceted {}",
+        ggsw_out.glwe_size().to_glwe_dimension().0,
+        fpksk_output_glwe_key_dimension.0
+    );
+
+    debug_assert!(
+        ggsw_out.glwe_size().0 * ggsw_out.decomposition_level_count().0
+            == fpksk_list.fpksk_count().0,
+        "The input vector of fpksk needs to have {} (ggsw.glwe_size * \
+        ggsw.decomposition_level_count) elements got {}",
+        ggsw_out.glwe_size().0 * ggsw_out.decomposition_level_count().0,
+        fpksk_list.fpksk_count().0,
+    );
+
+    // Output for every bootstrapping
+    let mut lwe_out_bs_buffer: LweCiphertext<Vec<Scalar>> = LweCiphertext::allocate(
+        Scalar::ZERO,
+        LweDimension(bsk_glwe_dimension.0 * bsk_polynomial_size.0).to_lwe_size(),
+    );
+    // Output for every pfksk that that come from the output GGSW
+    let mut glwe_out_pfksk_buffer = ggsw_out.as_mut_glwe_list();
+
+    let mut out_pfksk_buffer_iter = glwe_out_pfksk_buffer.ciphertext_iter_mut();
+
+    for level_idx in 0..level_cbs.0 {
+        homomorphic_shift_binary(
+            fourier_bsk,
+            &mut lwe_out_bs_buffer,
+            lwe_in,
+            buffers,
+            DecompositionLevelCount(level_idx + 1),
+            base_log_cbs,
+            delta_log,
+        );
+
+        for pfksk in fpksk_list.fpksk_iter() {
+            let mut glwe_out = out_pfksk_buffer_iter.next().unwrap();
+            pfksk.private_functional_keyswitch_ciphertext(&mut glwe_out, &lwe_out_bs_buffer);
+        }
+    }
+}
+
+/// Homomorphic shift for LWE without padding bit
+///
+/// Starts by shifting the message bit at bit #delta_log to the padding bit and then shifts it to
+/// the right by base_log * level.
+pub fn homomorphic_shift_binary<Scalar, C1, C2, C3>(
+    fourier_bsk: &FourierBootstrapKey<C1, Scalar>,
+    lwe_out: &mut LweCiphertext<C2>,
+    lwe_in: &LweCiphertext<C3>,
+    buffers: &mut FourierBuffers<Scalar>,
+    level_count_cbs: DecompositionLevelCount,
+    base_log_cbs: DecompositionBaseLog,
+    delta_log: DeltaLog,
+) where
+    Scalar: UnsignedTorus,
+    FourierBootstrapKey<C1, Scalar>: AsRefTensor<Element = Complex64>,
+    LweCiphertext<C2>: AsMutTensor<Element = Scalar>,
+    LweCiphertext<C3>: AsRefTensor<Element = Scalar>,
+{
+    let ciphertext_n_bits = Scalar::BITS;
+    let lwe_in_size = lwe_in.lwe_size();
+    let polynomial_size = fourier_bsk.polynomial_size();
+
+    let mut lwe_left_shift_buffer = LweCiphertext::allocate(Scalar::ZERO, lwe_in_size);
+    // Shift message LSB on padding bit, at this point we expect to have messages with only 1 bit
+    // of information
+    lwe_left_shift_buffer.fill_with_scalar_mul(
+        lwe_in,
+        &Cleartext(Scalar::ONE << (ciphertext_n_bits - delta_log.0 - 1)),
+    );
+
+    // Add q/4 to center the error while computing a negacyclic LUT
+    let mut shift_buffer_body = lwe_left_shift_buffer.get_mut_body();
+    shift_buffer_body.0 = shift_buffer_body
+        .0
+        .wrapping_add(Scalar::ONE << (ciphertext_n_bits - 2));
+
+    let mut pbs_accumulator =
+        GlweCiphertext::allocate(Scalar::ZERO, polynomial_size, fourier_bsk.glwe_size());
+
+    // Fill lut (equivalent to trivial encryption as mask is 0s)
+    // The LUT is filled with -alpha in each coefficient where
+    // alpha = 2^{log(q) - 1 - base_log * level}
+    for poly_coeff in pbs_accumulator
+        .get_mut_body()
+        .as_mut_polynomial()
+        .coefficient_iter_mut()
+    {
+        *poly_coeff = Scalar::ZERO.wrapping_sub(
+            Scalar::ONE << (ciphertext_n_bits - 1 - base_log_cbs.0 * level_count_cbs.0),
+        );
+    }
+
+    // Applying a negacyclic LUT on a ciphertext with one bit of message in the MSB and no bit
+    // of padding
+    fourier_bsk.bootstrap(lwe_out, &lwe_left_shift_buffer, &pbs_accumulator, buffers);
+
+    // Add alpha where alpha = 2^{log(q) - 1 - base_log * level}
+    // To end up with an encryption of 0 if the message bit was 0 and 1 in the other case
+    let out_body = lwe_out.get_mut_body();
+    out_body.0 = out_body
+        .0
+        .wrapping_add(Scalar::ONE << (ciphertext_n_bits - 1 - base_log_cbs.0 * level_count_cbs.0));
 }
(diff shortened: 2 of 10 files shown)


IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
```diff
@@ -43,7 +43,7 @@ pub struct LweCiphertextVectorGlweCiphertextDiscardingPrivateFunctionalPackingKe
     pub output_polynomial_size: PolynomialSize,
     pub decomposition_level: DecompositionLevelCount,
     pub decomposition_base_log: DecompositionBaseLog,
-    pub function_lipschitz_bound: usize,
+    pub function_log_scalar: usize,
```
agnesLeroy, line 46: What is this function log scalar?

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
```diff
@@ -86,21 +87,25 @@ pub trait PrivateFunctionalPackingKeyswitchKeyCreationEngine<
     OutputSecretKey,
     PrivateFunctionalPackingKeyswitchKey,
     CleartextVector,
+    FunctionScalarType,
```
agnesLeroy, line 90: What is FunctionScalarType here? It deserves an explanation in the Formal definition I think :slightly_smiling_face:

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
```diff
@@ -86,21 +87,25 @@ pub trait PrivateFunctionalPackingKeyswitchKeyCreationEngine<
...
     fn create_private_functional_packing_keyswitch_key(
         &mut self,
         input_key: &InputSecretKey,
         output_key: &OutputSecretKey,
         decomposition_level_count: DecompositionLevelCount,
         decomposition_base_log: DecompositionBaseLog,
         noise: StandardDev,
-        polynomial_scalar: &CleartextVector,
+        f: &dyn Fn(FunctionScalarType) -> FunctionScalarType,
+        polynomial: &CleartextVector,
```
agnesLeroy, line 108: Here I'm not sure we want to use a CleartextVector, since cleartexts are meant to be only objects that we plan to encode & encrypt. I think we'd need a type defined in concrete-commons rather, can you confirm @aPere3?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
```diff
@@ -121,6 +132,179 @@ where
...
+
+/// Circuit bootstrapping for binary messages, i.e. containing only one bit of message
+pub fn circuit_bootstrap_binary<Scalar>(
+    fourier_bsk: &FourierBootstrapKey<AlignedVec<Complex64>, Scalar>,
+    lwe_in: &LweCiphertext<Vec<Scalar>>,
+    buffers: &mut FourierBuffers<Scalar>,
+    level_cbs: DecompositionLevelCount,
+    base_log_cbs: DecompositionBaseLog,
+    delta_log: DeltaLog,
+    vec_fpksk: &[PrivateFunctionalPackingKeyswitchKey<Vec<Scalar>>],
```
agnesLeroy, line 145: It's a bit strange to name it vec_ when it's a slice no? Maybe we could name it fpksks?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
```diff
@@ -121,6 +132,179 @@ where
...
+/// Homomorphic shift for LWE without padding bit
+///
+/// Starts by shifting the message bit at bit #delta_log to the padding bit and then shifts it to
+/// the right by base_log * level.
+pub fn homomorphic_shift_binary<Scalar, C1, C2>(
+    fourier_bsk: &FourierBootstrapKey<AlignedVec<Complex64>, Scalar>,
+    lwe_out: &mut LweCiphertext<C1>,
+    lwe_in: &LweCiphertext<C2>,
+    buffers: &mut FourierBuffers<Scalar>,
+    level_cbs: DecompositionLevelCount,
```
agnesLeroy, line 254: Maybe we could name this one level_count_cbs?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
```diff
@@ -121,6 +132,179 @@ where
...
+        .wrapping_add(Scalar::ONE << (ciphertext_n_bits - 2));
+
+    let mut pbs_accumulator =
+        GlweCiphertext::allocate(Scalar::ZERO, polynomial_size, fourier_bsk.glwe_size());
+
+    // Fill lut (equivalent to trivial encryption as mask is 0s)
+    // The LUT is filled with -alpha in each coefficient where
+    // alpha = delta * 2 ^ {-level * base_log} where delta = q / 2 = 2 ^ 63 for a 64 bits ciphertext
+    // alpha = delta / 2 ^ {level * base_log}
+    // alpha = delta / (base ^ level)
```
agnesLeroy, line 287: I don't understand this comment: why are there three alpha definitions?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
```diff
@@ -121,6 +132,179 @@ where
...
+
+    // Applying a negacyclic LUT on a ciphertext with one bit of message in the MSB and no bit
+    // of padding
+    fourier_bsk.bootstrap(lwe_out, &lwe_left_shift_buffer, &pbs_accumulator, buffers);
 
-    output_lwe_list
+    // Add alpha where
+    // alpha = delta * 2 ^ {-level * base_log} where delta = q / 2 = 2 ^ 63 for a 64 bits ciphertext
+    // alpha = delta / 2 ^ {level * base_log}
+    // alpha = delta / (base ^ level)
```
agnesLeroy, line 304: Same question here

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
```diff
@@ -140,3 +153,245 @@ pub fn test_extract_bits() {
         }
     }
 }
+
+// test the CBS with private functional ks
```
agnesLeroy, line 157: Maybe we could name the test test_circuit_bootstrap_binary? I'd rather we write circuit bootstrap instead of cbs where we can :slightly_smiling_face: I do not understand the comment about what the test does: maybe @Loris-B you can help us come up with a description?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
```diff
@@ -111,8 +116,16 @@ pub fn test_extract_bits() {
         // Bit extraction
         // Extract all the bits
         let number_values_to_extract = ExtractedBitsCount(64 - delta_log.0);
-        let result_lwe = extract_bits(
+
+        let mut lwe_out_list = LweList::allocate(
```
agnesLeroy, line 120: Why did you choose to take the allocation out of the function? Should we do the same for the cbs?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
```diff
@@ -140,3 +153,245 @@ pub fn test_extract_bits() {
...
+
+    assert_eq!(
+        first_value_decrypted_ext_prod_rounded >> delta_log.0,
+        (first_value_decrypted_input_ext_prod_rounded >> delta_log.0) * value
+    );
+}
+
+/////////////////////////////////////////////////////////////
+////////////////// for param creation ///////////////////////
+/////////////////////////////////////////////////////////////
```
agnesLeroy, line 329: I think we can remove those comment lines no?

agnesLeroy (commented):
Hi @IceTDrinker! Thanks a lot for this PR: here come some questions/suggestions. I do not know the algorithm of the circuit bootstrap so it's not possible for me to tell whether the crypto is correct (or even if it could be optimized) :disappointed: I'll let @Loris-B and/or @jborfila give their review on this :slightly_smiling_face:

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: sorry those three lines are equivalent

<=>

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
IceTDrinker: hmm, I might have made a mistake trying to get back the old behavior from Carl's PR, I'll check

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
agnesLeroy: Ahah I hadn't even paid attention :sweat_smile:

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
IceTDrinker: Well it's just to specify on what kind of data the function we apply during the creation of the key operates, it needed to be specified as a type parameter

How should it appear in the formal def ? Or maybe I forgot to update the formal def to its previous state

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: sure

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
IceTDrinker: This may be a comment I did not go over, I'm not sure anymore 😅

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
IceTDrinker: I do that in steps, maybe I shouldn't, and then I update to fit the behavior of a discarding operation

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
IceTDrinker: yes maybe I need to go over the test code, maybe I wen too fast but this way we have the PR with the algorithm available for review

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: I think I did not add the symbol as it was going over the comment line length limit

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: well rust complains when we pass a ref to a vec because taking a slice allows to manage more containers/cases as long as they can be coerced into a slice, so the name here though a bit misleading does not shock me

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
IceTDrinker: so the previous behavior in the fixture when using a function (which I reworked because it was easier to use) used this function_log_scalar to build a power of two to be applied by passing a closure as you can see in the diff for the fixture.

Not sure how it could be renamed though

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
agnesLeroy: I thought the engine would take a polynomial & a reference to a function now. Actually there is no formal definition written yet sorry :disappointed:

IceTDrinker: Ah I think I understand what you mean now, I can change it to an `Fn` type if you want but it's more annoying to write and error prone if someone chooses a weird fn type

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
agnesLeroy: ok :slightly_smiling_face:

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
IceTDrinker: So I made the choice to force the function to be an `Fn` and just specify which type goes and goes out of that function

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
IceTDrinker: done

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
IceTDrinker: Updated, let me know if it's better

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: done

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
agnesLeroy: Oh ok, my bad my comments are very unclear :sweat_smile: Makes sense to me now: you're restraining the data type the function can operate on on u32 and u64.

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
agnesLeroy: Actually I really think we should have a type for the polynomial other than CleartextVector, since cleartexts can be many other things than u32/u64.

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
```diff
@@ -121,6 +132,180 @@ where
...
+        base_log_cbs,
+    )
+}
+
+/// Homomorphic shift for LWE without padding bit
+///
+/// Starts by shifting the message bit at bit #delta_log to the padding bit and then shifts it to
+/// the right by base_log * level.
+pub fn homomorphic_shift_binary<Scalar, C1, C2>(
+    fourier_bsk: &FourierBootstrapKey<AlignedVec<Complex64>, Scalar>,
```
aPere3, line 250: Nittpicking, because we will rework that soon, but here, there is no need to fix it to be an alignedvec, I believe.

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
```diff
@@ -121,6 +132,180 @@ where
...
+            .iter()
+            .all(|fpksks| fpksks.output_glwe_key_dimension() == fpksk_output_glwe_key_dimension),
+        "Not all fpksk from vec_fpksk have the same output_glwe_key_dimension"
+    );
+
+    // output for every bootstrapping
+    let mut lwe_out_bs_buffer: LweCiphertext<Vec<Scalar>> = LweCiphertext::allocate(
+        Scalar::ZERO,
+        LweDimension(bsk_glwe_dimension.0 * bsk_polynomial_size.0).to_lwe_size(),
+    );
```
aPere3, line 202: Any reason not to pass it as argument to the function ?

aPere3 (commented):
Hello @IceTDrinker ,

Thanks for this PR ! 

I have a few questions I'd like to ask !

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
```diff
@@ -86,21 +87,25 @@ pub trait PrivateFunctionalPackingKeyswitchKeyCreationEngine<
     OutputSecretKey,
     PrivateFunctionalPackingKeyswitchKey,
     CleartextVector,
+    FunctionScalarType,
 >: AbstractEngine where
     InputSecretKey: LweSecretKeyEntity,
     OutputSecretKey: GlweSecretKeyEntity,
     CleartextVector: CleartextVectorEntity,
     PrivateFunctionalPackingKeyswitchKey: PrivateFunctionalPackingKeyswitchKeyEntity,
+    FunctionScalarType: UnsignedInteger,
```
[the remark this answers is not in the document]
IceTDrinker: I'm not sure, probably not now that you mention it :|

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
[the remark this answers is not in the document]
IceTDrinker: We could add one, I was not the one to implement the PFKSK so I'm not sure what is behind that name

@jborfila @Loris-B would know maybe ?

concrete-core/src/commons/crypto/glwe/keyswitch.rs
```diff
@@ -917,8 +917,6 @@ impl<Cont> PrivateFunctionalPackingKeyswitchKey<Cont> {
 
     /// Fills the current private functional keyswitch key container with an actual private
     /// functional keyswitching key constructed from an input and an output key.
-    /// Note that the private function is determined as x -> pol * x is where `pol` is a fixed
-    /// polynomial scalar.
```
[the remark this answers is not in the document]
IceTDrinker: actually we do, but as the function is not determined this way anymore this comment was out of date

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: hmm, I can check, but if the container is not an aligned vec it would not work for bootstrapping with fftw ?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: I do that in steps, but it seems it is confusing people more than anything, so I'll see to make it an input parameter

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
IceTDrinker: what do you mean about the precision to encode integers ?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: I think this may have been done for the sake of going a bit faster (not having the list type)

Also there is a List of polynomial list down the line for the vertical packing so :/

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: Seems to work ok, I'll update the various parts of the file

Loris-B (commented):
thank, nice job!

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
```diff
@@ -121,6 +132,180 @@ where
...
+
+    debug_assert!(
+        vec_fpksk
+            .iter()
+            .all(|fpksks| fpksks.input_lwe_key_dimension() == fpksk_input_lwe_key_dimension),
+        "Not all fpksk from vec_fpksk have the same input_lwe_key_dimension"
+    );
+
+    debug_assert!(
+        fpksk_input_lwe_key_dimension.0 == bsk_polynomial_size.0,
```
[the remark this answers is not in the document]
IceTDrinker: thanks !

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
```diff
@@ -121,6 +132,180 @@ where
...
+    );
+    // output for every pfksk that will eventually be outputed in a GGSW
+    let mut glwe_out_pfksk_buffer = GlweList::allocate(
+        Scalar::ZERO,
+        fpksk_output_polynomial_size,
+        fpksk_output_glwe_key_dimension,
+        CiphertextCount(level_cbs.0 * vec_fpksk.len()),
+    );
+
+    {
```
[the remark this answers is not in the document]
IceTDrinker: I think the way I use the iterator means there is a borrow, scoping the borrow means I can borrow again from the object after the scope, I'll check and add a comment

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
```diff
@@ -140,3 +153,241 @@ pub fn test_extract_bits() {
...
+        decomposer.closest_representable(first_value_decrypted_input_ext_prod);
+
+    assert_eq!(
+        first_value_decrypted_ext_prod_rounded >> delta_log.0,
+        (first_value_decrypted_input_ext_prod_rounded >> delta_log.0) * value
+    );
+}
+
+/// Helper function for circuit bootstrapping test to create fpksk
+pub fn create_vec_pfksk<LWEKeyCont, GLWEKeyCont, Scalar>(
```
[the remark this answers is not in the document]
IceTDrinker: definitely, if you think it's a function you use often !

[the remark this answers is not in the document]
IceTDrinker: I've seen it in tests mostly, is it something you use elsewhere in concrete if you can remember ?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: fine by me !

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: Should I create the corresponding private struct and public entity ?

As for the list of list of polynomials I'm going to say we probably need to rethink this a little :|

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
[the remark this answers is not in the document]
Loris-B: We use it every time we need to do a CBS (so every time we need to do a WoPBS)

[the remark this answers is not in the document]
IceTDrinker: yes then it's something we should definitely have in the API !

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: done

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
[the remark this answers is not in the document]
IceTDrinker: added the comment, it's because of a mutable borrow

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
aPere3: Sorry for being unclear. What I meant is that what would be best is to have in the commons something like:
```rust
// in src/commons
pub struct PublicInteger(u32); // Or u16 ? or u64 ?
```

That would, for instance, make the `LweCiphertextCleartextDiscardingMultiplicationEngine` look like this (the name should probably be changed also):
```rust
pub trait LweCiphertextCleartextDiscardingMultiplicationEngine<
    InputCiphertext,
    OutputCiphertext,
>: AbstractEngine where
    InputCiphertext: LweCiphertextEntity,
    OutputCiphertext: LweCiphertextEntity,
{
    /// Multiply an LWE ciphertext with a cleartext.
    fn discard_mul_lwe_ciphertext_cleartext(
        &mut self,
        output: &mut OutputCiphertext,
        input_1: &InputCiphertext,
        input_2: &PublicInteger,
    ) -> Result<(), LweCiphertextCleartextDiscardingMultiplicationError<Self::EngineError>>;

    /// Unsafely multiply an LWE ciphertext with a cleartext.
    ///
    /// # Safety
    /// For the _general_ safety concerns regarding this operation, refer to the different variants
    /// of [`LweCiphertextCleartextDiscardingMultiplicationError`]. For safety concerns _specific_
    /// to an engine, refer to the implementer safety section.
    unsafe fn discard_mul_lwe_ciphertext_cleartext_unchecked(
        &mut self,
        output: &mut OutputCiphertext,
        input_1: &InputCiphertext,
        input_2: &PublicInteger,
    );
}
```
Now the thing is that in the implementation with `LweCiphertext32` we are directly good to go, as both use u32, but with u64, we have to cast it. So the question we should answer is: what is the range of values that this public integer can take, and does it change with ciphertext precision.

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
aPere3: Hmm for the list of polynomials, I think it should be enough to pass a list of polynomials, which we then chunk. I think we already have that in the code.

For the rest, I would say yes. The usual way would be to create a new structure. No need to put something fancy, a struct with iterators would be enough I believe.

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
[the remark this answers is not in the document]
aPere3: Hmm. I did not expect that we would have to make this trait part of the public api ... That is unfortunate :/ It is unclear to me if we can avoid that for now ...

concrete-core/src/commons/crypto/glwe/keyswitch.rs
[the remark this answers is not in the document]
aPere3: Is the polynomial used in the function now ?

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
aPere3: ok cool !

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
IceTDrinker: can we maybe remove the bound and then just use u32/u64 ?

I don't think this pattern is going to simplify our lives

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: Yep for the polynomial list with chunks I though about it afterwards

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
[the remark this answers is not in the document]
IceTDrinker: This is very fixture specific it seems, as the Fn passed to compute the mutiplicative factor will most likely capture a scalar (so a u32 or a u64 depending on the case), so I don't think we can/should add a type for it

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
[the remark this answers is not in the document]
IceTDrinker: We can remove this bound if it's an issue, the rest constrains it naturally

concrete-core/src/commons/crypto/glwe/keyswitch.rs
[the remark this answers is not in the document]
IceTDrinker: the function is used to compute a multiplication factor by which a key bit is multiplied and then the polynomial passed as input is multiplied by that as well before storing the result in the FPKSK

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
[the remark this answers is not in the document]
IceTDrinker: what should this function be named ? it does a very specific preparation for the vector of FPKSK

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
agnesLeroy: Ok, it's really specific to the fixture so let's leave it like this

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
agnesLeroy: Let me try to summarize: when we pass parameters to engines, either they have to:
- Have a generic type (u32/u64 usually), in which case what we did until now was to create entities (like cleartexts) for them.
- Or, they have a simple type and we can just define a type in commons to wrap it (we do it systematically otherwise there are just too many variables wih the same type and it's error-prone).

In the present case the polynomial we have to provide as an input is an array of U32 or U64 values. Then for me the question is: can we give a U64 polynomial even if the ciphertext is represented with U32? Even if it's possible, it seems a bit complicated actually, so if we say we pass a U32 polynomial with a U32 ciphertext specifically, I would create a new entity for the polynomial and engines to create/retrieve it.

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_glwe_ciphertext_discarding_private_functional_packing_keyswitch.rs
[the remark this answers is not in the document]
agnesLeroy: Agreed

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
agnesLeroy: The test name looks good thanks. @Loris-B could you send here a more detailed description of what the test does? Just reading the sources is not enough for me to understand the details, we'll need more comments. :)

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
IceTDrinker: I'm fine with creating new entities, they are all glorified wrappers over vec anyways :) (which have their value for typing of course)

concrete-core/src/commons/crypto/glwe/keyswitch.rs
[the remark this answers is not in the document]
IceTDrinker: the polynomial is actually a part of a glwe secret key in the way the fpksk is used so we can't really make without that polynomial

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: done, we now fill an output ggsw directly, the temp buffer is kept as is, if we see that this allocation costs too much we can create variants taking that buffer as an input

concrete-core/src/specification/engines/lwe_private_functional_packing_keyswitch_key_creation.rs
IceTDrinker: As discussed we are keeping the cleatext vector for now

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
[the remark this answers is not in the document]
IceTDrinker: function moved to fill_with_fpksk_for_binary_cbs in the PrivateFunctionalPackingKeyswitchKeyList feedback welcome ! @Loris-B

concrete-core/src/backends/fftw/private/crypto/wop_pbs/mod.rs
IceTDrinker: Added PrivateFunctionalPackingKeyswitchKeyList  with a function to fill_with_fpksk_for_binary_cbs not sure about the name yet, I pinged Loris to see if there is another name describing how the list is filled that might not be directly linked to the circuit bootstrap

IceTDrinker: Public API will come for thie PrivateFunctionalPackingKeyswitchKeyList once it is required so basically when the public API for the circuit bootstrap will be done (as I think the WoP PBS call can hide some of those details)

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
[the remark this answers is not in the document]
IceTDrinker: naming checked with JB, it is now fill_with_fpksk_for_circuit_bootstrap

IceTDrinker:
I think this is close to merging last questions remaining for now :

- FunctionScalar -> remove the integer bound in the Public API as it would be enforced by the other traits ? Or "leak" the private UnsignedInteger @aPere3 let me know
- missing a description potentially for the test "test_circuit_bootstrapping_binary" @Loris-B if you can take a look at @agnesLeroy question here: https://github.com/zama-ai/concrete-core/pull/130#discussion_r953599819

concrete-core/src/backends/fftw/private/crypto/wop_pbs/test.rs
IceTDrinker: Test updated to check the content of the GGSW directly

aPere3:
@IceTDrinker -> Better not to leak `UnsignedInteger`. Let's leave it unbounded for now.

IceTDrinker:
> @IceTDrinker -> Better not to leak `UnsignedInteger`. Let's leave it unbounded for now.

If it looks ok can you approve this version of the PR ? I need to push a commit to remove a forgotten comment with a TODO from the tests

aPere3 (approved):

agnesLeroy (approved):

PR #138, opened 2022-08-22
feat(core): add PartialOrd, Ord derives to [Glwe/Lwe][Dimension/Size]

tmontaigu:
### Description

This adds PartialOrd, Ord derives for GlweSize, LweSize, GlweDimension, GlweSize so that they can be used on types
which can then derive PartialOrd, Ord to be used as Keys for BTreeMap.

### Checklist 

### DIFF
--- a/concrete-commons/src/parameters.rs
+++ b/concrete-commons/src/parameters.rs
@@ -59,7 +59,7 @@ pub struct GswCiphertextCount(pub usize);
 pub struct GgswCiphertextCount(pub usize);
 
 /// The number of scalars in an LWE ciphertext, i.e. the number of scalar in an LWE mask plus one.
-#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 #[cfg_attr(feature = "serde_serialize", derive(Serialize, Deserialize))]
 pub struct LweSize(pub usize);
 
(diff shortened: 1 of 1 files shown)


tmontaigu:
a test like this ?

```Rust

#[cfg(test)]
mod test {

/// This will will to compile if any of the member is not Partial{Ord, Eq} or Eq or Ord
#[derive(PartialOrd, Ord, PartialEq, Eq)]
#[allow(dead_code)]
struct AssertIsOr {
 _lwe_size: LweSize,
 /// ...
}
```

Where should it be placed ?

IceTDrinker:
> a test like this ?
> 
> ```rust
> #[cfg(test)]
> mod test {
> 
> /// This will will to compile if any of the member is not Partial{Ord, Eq} or Eq or Ord
> #[derive(PartialOrd, Ord, PartialEq, Eq)]
> #[allow(dead_code)]
> struct AssertIsOrd {
>  _lwe_size: LweSize,
>  /// ...
> }
> ```
> 
> Where should it be placed ?

It was a suggestion (I don't know how commons is structured), if there is no current place in commons then I guess we could make one

agnesLeroy:
Currently there are no tests in concrete-commons, but we plan to merge its contents into concrete-core (then it'll be rather straightforward to add tests, it won't imply changing the CI).

IceTDrinker:
Then @tmontaigu don't bother, we have your proposal for the Assert struct 🙂

agnesLeroy:
@slab-ci cpu_test

PR #139, opened 2022-08-24
Update the Optalysys simulator version and include recent changes to concrete-core

FlorentCLMichel:
### Description

This PR has two aims: 

* Include some of the recent changes from the main branch of concrete-core.
* Update the Optalysys OFT simulator version from 2 to 3b (about 10x faster). 

This version should be up to date with the main branch as of 19 August 2022. Minor changes were made to the Optalysys backend in line with the new structure (notably the fact that some files were moved from commons to the fftw backend, and thus not accessible to the Optalysys backend any more). 

The new Optalysys simulator is functionally similar to the previous one but provides a higher-level simulation, making it significantly faster at the cost of being slightly more pessimistic when estimating the runtime on optical hardware (as a few hardware-specific optimisations can't be reflected in the higher-level simulation; in practice, the difference is generally negligible for power-of-2 Fourier transforms as used in FHE). It also provides faster ‘disordered’ direct and inverse Fourier transform functions with one less butterfly step. These functions are not yet used in the backend as they would break some of the existing tests, but will be used in the upcoming hardware implementation.


agnesLeroy:
Hi @FlorentCLMichel! I think here we want to merge only the last commit on your branch: I don't know why the 159 other ones are included in the merge :thinking: Do you have an idea? Is it because you rebased the branch on top of main maybe?

FlorentCLMichel:
Hi @agnesLeroy ! Thank you for having looked at this! yes, I think most of the 159 commits are from the main branch (I fetched the main branch on  mine before doing the PR to include the recent changed to main). 

> What I can do is rebase the feature/optalysys_backend branch on top of main, and then you can open a PR just for your last commit: how does it sound to you?

Sounds good to me! Shall I close this PR and open a new one once you have done the rebase?

agnesLeroy:
@FlorentCLMichel I have just pushed the rebase to feature/optalysys_backend: you can force push onto your fork branch to directly update this PR, no need to create a new one. You'll have to:
- fetch feature/optalysys_backend in your local fork
- reset your optalysys_august_2022 onto it
- cherry-pick commit 006ca0bdaeef6c4456c208a05937b67d9d291c3a (you may have conflicts due to changes in the last commit of feature/optalysys_backend)
- force push onto your fork on github
Let me know if any of this is unclear! :slightly_smiling_face:

FlorentCLMichel:
@agnesLeroy Fantastic! Thanks a lot for having done the rebase so quickly and for the explanation! I will try to do that and let you know if I have issues.

agnesLeroy:
@FlorentCLMichel there has been a problem it seems: many changes in the commit are reverts of changes we made in main :disappointed:

FlorentCLMichel:
@agnesLeroy Indeed - I think the reason is that I did not use the `--hard` option when resetting. I'm re-trying with this option; hopefully it will solve the issue! 
(Converting this PR to draft until this is solved.)

FlorentCLMichel:
Hopefully this should be resolved :)  Thanks again for your help @agnesLeroy ! Please let me know if anything else needs to be changed.

concrete-core/src/backends/optalysys/implementation/engines/destruction.rs
```diff
@@ -0,0 +1,29 @@
+use crate::backends::optalysys::implementation::engines::OptalysysEngine;
```
agnesLeroy, line 1: Hi @FlorentCLMichel! This file can actually be removed: we no longer have destroy engines and all backends should implement Drop for automatic memory management. In your case you have nothing to do yet since the simulator takes care of freeing memory itself.

FlorentCLMichel: Hi @agnesLeroy ! Thanks for letting me know! I have just updated the PR to remove this file.

agnesLeroy (commented):
Hi again @FlorentCLMichel! Looking at the PR again I'm wondering whether we couldn't put the twiddles file in commons rather than duplicating it for the two backends? Also, there's a cuda_script added in the PR that's been mistakenly added. :smiley: Otherwise everything looks great!

FlorentCLMichel:
> Also, there's a cuda_script added in the PR that's been mistakenly added. 

Well spotted! Not sure why this file appeared - I had manually deleted it but it seems it had crept up again. I've just updated the PR to remove it.

> Looking at the PR again I'm wondering whether we couldn't put the twiddles file in commons rather than duplicating it for the two backends? 

That makes sense to me. I think it was in`commons in the previous version, but has been moved to `fftw::private::math` since then. Do you know if there was a particular reason for this change? (Maybe to avoid having something in commons which is not used by the other backends?) I'm happy to move it back to commons if you think it is better.

agnesLeroy:
@FlorentCLMichel I think it was moved in the fftw backend since it was used only there. Let's put it back in commons for now, sorry we're changing things over again :sweat_smile:

FlorentCLMichel:
@agnesLeroy No problem at all!

agnesLeroy (approved):

PR #144, opened 2022-08-26
feat(concrete_npe): remove torus type generics

mayeul-zama:
### Description

Replace remove torus type generics by a function parameter log2_modulus in all relevant functions in concrete-npe.

- the generics torus type in only ever used to extract the number of bits, so we don't loose any useful information by only giving the log2_modulus
- we only ask for the information we need so it's less constraining for callers
- the generics are contaminating, which forces us to use generics too in the optimizer (log2_modulus -> generic type) is not possible (or very ugly: match log2_modulus {32 => f::<u32>(...), 64 => f::<u64>(...))
- for another client of concrete-npe using generics, it's possible to get the log2_modulus from the generic type (genric_type -> log2_modulus) is clean (f(..., T::BITS))
- we add a function parameter but we also remove a generic type, so we slightly reduce signature complexity
- with generics, monomorphisation is done (for each type we call generic a function with, the compiler creates a separate version) which may increase compile times and binary size
- theoretical future advantage: with the log2_modulus, we can call the functions for any number of bits, which may be useful for a FPGA implementation which doesn't use 16, 32, 64 or 128 bits

### Checklist 

### DIFF
--- a/concrete-commons/src/dispersion.rs
+++ b/concrete-commons/src/dispersion.rs
@@ -16,8 +16,6 @@
 #[cfg(feature = "serde_serialize")]
 use serde::{Deserialize, Serialize};
 
-use crate::numeric::UnsignedInteger;
-
 /// A trait for types representing distribution parameters, for a given unsigned integer type.
 //  Warning:
 //  DispersionParameter type should ONLY wrap a single native type.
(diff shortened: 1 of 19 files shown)


mayeul-zama:
@slab-ci cpu_test

mayeul-zama:
@slab-ci cpu_test

concrete-commons/src/dispersion.rs
```diff
@@ -32,17 +30,13 @@ pub trait DispersionParameter: Copy {
     /// $\log\_2(\sigma)=p$
     fn get_log_standard_dev(&self) -> f64;
     /// For a `Uint` type representing $\mathbb{Z}/2^q\mathbb{Z}$, we return $2^{q-p}$.
-    fn get_modular_standard_dev<Uint>(&self) -> f64
-    where
-        Uint: UnsignedInteger;
+    fn get_modular_standard_dev(&self, log2_modulus: u64) -> f64;
```
agnesLeroy, line 33: Just a small thing: we could use a u32 or even a u8 for log2_modulus I guess, it's always going to be 32 or 64...

IceTDrinker: I think rust uses u32 for these like when calling e.g. `2u64.pow(/* u32 here */)`

PR #147, opened 2022-08-30
chore(crates): change the way to declare deps to be git friendly

IceTDrinker:
### Description

- we use multiple location dep declaration, path is used for local builds
or for builds using the git repo
- the version is used for crates published on crates.io

### Checklist 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,13 +14,6 @@ members = [
     "concrete-cuda",
 ]
 
-[patch.crates-io]
-concrete-npe = { path = "concrete-npe" }
-concrete-core = { path = "concrete-core" }
-concrete-csprng = { path = "concrete-csprng" }
-concrete-commons = { path = "concrete-commons" }
-concrete-cuda = { path = "concrete-cuda" }
-
 [profile.release]
 lto = "fat"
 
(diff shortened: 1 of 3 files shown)


IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci gpu_test

agnesLeroy (approved):
Looks good to me, anyway we'll be moving concrete-commons inside concrete-core shortly. Then it'll mean we have to bump concrete-npe version between crates.io pushes when there are changes to it we need in concrete-core.

agnesLeroy:
@aPere3 I'm assigning you as reviewer as well, since we've been through this so many times :sweat_smile:

IceTDrinker:
It seems to be working for Mayeul's use case, and it is apparently a common pattern in the rust projects that are multicrates.

Relevant infos from the docs :

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

```
It is possible to specify both a registry version and a git or path location. The git or path dependency will be used locally (in which case the version is checked against the local copy), and when published to a registry like crates.io, it will use the registry version. Other combinations are not allowed.
```

PR #148, opened 2022-08-30
feat(npe): add noise propagation estimates for WoP-PBS

Carl-Zama:
### Resolves:
https://github.com/zama-ai/concrete-core-internal/issues/316
https://github.com/zama-ai/concrete-core-internal/issues/317
https://github.com/zama-ai/concrete-core-internal/issues/318
https://github.com/zama-ai/concrete-core-internal/issues/319

### Description

This PR adds the noise propagation estimates for the WoP-PBS and its constituent parts: bit extraction, circuit bootstrapping and vertical packing.
This PR also updates the private functional packing keyswitch key noise propagation estimation function and the associated fixture to remove the torus type generics.

### Checklist 

### DIFF
--- a/concrete-npe/src/operators.rs
+++ b/concrete-npe/src/operators.rs
@@ -1,6 +1,7 @@
 use concrete_core::commons::numeric::{CastInto, UnsignedInteger};
 use concrete_core::prelude::{
-    DecompositionBaseLog, DecompositionLevelCount, GlweDimension, LweDimension, PolynomialSize,
+    BinaryKeyKind, DecompositionBaseLog, DecompositionLevelCount, ExtractedBitsCount,
+    GlweDimension, LweDimension, PolynomialSize,
 };
 /// Contains material needed to estimate the growth of the noise when performing homomorphic
 /// computation
@@ -534,7 +535,6 @@ where
 /// let dispersion_ks = Variance(2_f64.powi(-40));
 /// let function_lipschitz_bound = 10.;
 /// let var_ks = estimate_private_functional_keyswitch_noise_lwe_to_glwe_with_constant_terms::<
-///     u64,
 ///     _,
 ///     _,
 ///     BinaryKeyKind,
@@ -545,18 +545,19 @@ where
 ///     base_log,
 ///     l_ks,
 ///     function_lipschitz_bound,
+///     64,
 /// );
 /// ```
-pub fn estimate_private_functional_keyswitch_noise_lwe_to_glwe_with_constant_terms<T, D1, D2, K>(
+pub fn estimate_private_functional_keyswitch_noise_lwe_to_glwe_with_constant_terms<D1, D2, K>(
     lwe_mask_size: LweDimension,
     dispersion_lwe: D1,
     dispersion_ksk: D2,
     base_log: DecompositionBaseLog,
     level: DecompositionLevelCount,
     function_lipschitz_bound: f64,
+    log2_modulus: u32,
 ) -> Variance
 where
-    T: UnsignedInteger,
     D1: DispersionParameter,
     D2: DispersionParameter,
     K: KeyDispersion,
@@ -564,22 +565,22 @@ where
     let n = lwe_mask_size.0 as f64;
     let base = 2_f64.powi(base_log.0 as i32);
     let b2l = 2_f64.powi((base_log.0 * 2 * level.0) as i32);
-    let q_square = 2_f64.powi((2 * T::BITS as u32) as i32);
+    let q_square = 2_f64.powi((2 * log2_modulus) as i32);
     let r2 = square(function_lipschitz_bound);
 
     // res 1
-    let res_1 = r2 * dispersion_lwe.get_modular_variance(T::BITS as u32);
+    let res_1 = r2 * dispersion_lwe.get_modular_variance(log2_modulus);
 
     // res 2
     let res_2 = r2
         * n
         * (q_square / (12. * b2l) - 1. / 12.)
-        * (K::variance_key_coefficient(T::BITS as u32).get_modular_variance(T::BITS as u32)
+        * (K::variance_key_coefficient(log2_modulus).get_modular_variance(log2_modulus)
             + square(K::expectation_key_coefficient()));
 
     // res 3
-    let res_3 = r2 * n / 4.
-        * K::variance_key_coefficient(T::BITS as u32).get_modular_variance(T::BITS as u32);
+    let res_3 =
+        r2 * n / 4. * K::variance_key_coefficient(log2_modulus).get_modular_variance(log2_modulus);
 
     // res 4
     let res_4 = r2 * (q_square / (12. * b2l) - 1. / 12.);
@@ -587,11 +588,11 @@ where
     // res 5
     let res_5 = (n + 1.)
         * (level.0 as f64)
-        * dispersion_ksk.get_modular_variance(T::BITS as u32)
+        * dispersion_ksk.get_modular_variance(log2_modulus)
         * (square(base) + 2.)
         / 12.;
 
-    Variance::from_modular_variance(res_1 + res_2 + res_3 + res_4 + res_5, T::BITS as u32)
+    Variance::from_modular_variance(res_1 + res_2 + res_3 + res_4 + res_5, log2_modulus)
 }
 
 /// Computes the dispersion of the non-constant GLWE terms after an LWE to GLWE keyswitch.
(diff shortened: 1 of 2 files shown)


concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,295 @@ where
     Variance::from_modular_variance(res_1 + res_2, log2_modulus)
 }
 
+/// Computes the dispersions of ciphertexts encrypting the bits after bit after bit extraction,
```
IceTDrinker, line 861: "after bit after bit"

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,295 @@ where
...
+/// the dispersion of the LSB ciphertext is first and the MSB last.
```
IceTDrinker, line 862: could we mabe reverse the order ?

We are outputting ciphertexts from MSB to LSB in the algorithm we have in core

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,295 @@ where
...
+///     level_bsk,
+///     64,
+/// );
+/// ```
+#[allow(clippy::too_many_arguments)]
+pub fn estimate_bit_extraction_noise<D1, D2, D3, K1, K2>(
+    number_of_bits_to_extract: ExtractedBitsCount,
+    input_lwe_mask_size: LweDimension,
+    lwe_mask_size_after_ks: LweDimension,
+    rlwe_mask_size: GlweDimension,
```
IceTDrinker, line 905: rlwe/glwe ?

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,295 @@ where
...
+                log2_modulus,
+            );
+        }
+    }
+    output_dispersions
+}
+
+/// Computes the dispersion of a circuit bootstrapping for a binary message and a binary secret
+/// key, each ciphertext output is the result of a PBS followed by a private functional
+/// keyswitch where the private function is multiplication by a secret coefficient (or 1).
```
IceTDrinker, line 963: `where the private function is multiplication by a secret coefficient (or 1)`

looks like a word is missing , `the private function is a/the multiplication...` ?

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,295 @@ where
...
+            rlwe_mask_size,
+            base_log,
+            level,
+            dispersion_bsk,
+            log2_modulus,
+        ),
+        dispersion_ksk,
+        base_log,
+        level,
+        1_f64, // Binary (or ternary) keys mean the lipschitz bound is 1
```
IceTDrinker, line 1022: doesn't the secret scalar used influence that value ?

Though here it's true we only use 1 or -1 as a coefficient for pfksk built for a cbs

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,295 @@ where
...
+        log2_modulus,
+    )
+}
+
+/// Compute the dispersion after a vertical/hybrid packing. The output ciphertext is the result
```
IceTDrinker, line 1027: I think it's only a vertical packing we are doing in the end after discussing a bit more with Loris ?

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,295 @@ where
...
+/// of n sequential CMUX operations where n is the number of GGSW ciphertexts given.
+/// During a PBS the final blind rotation also performs n sequential CMUX operations so we can
+/// use the same noise estimation formula.
+/// # Example
```
IceTDrinker, line 1031: Just to chime in on this (I'm probably way out of my depth in terms of maths) but it does not seem to fit what I saw in the algorithm: we are doing a cmux tree and then a blind rotate with some potential leftover ggsw in the vertical packing, so is this assumption correct ?

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,295 @@ where
...
+///     rlwe_mask_size,
+///     base_log_cb,
+///     level_cb,
+///     dispersion_cb_bsk,
+///     dispersion_cb_pfksk,
+///     64,
+/// );
+/// ```
+#[allow(clippy::too_many_arguments)]
+pub fn estimate_wop_pbs_noise<D1, D2, K1, K2>(
```
IceTDrinker, line 1113: I'm not sure about whether this one is needed in the end as concrete will build it's own wop PBS from the primitives we have and the compiler will do the same

Also a wop PBS currently starts by a bit extract so I guess we would be missing something here potentially ?

IceTDrinker (commented):
Some general comments/question I'm not good enough in terms of FHE maths and noise estimation to make accurate maths remarks, I just tried to make what you write match with what I saw while getting the algorithms in core

IceTDrinker:
test distribution failure unrelated to your pr, relaunching

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
Looks like a doc test for one of the noise formulas was not updated for the log2_modulus change

concrete-npe/src/operators.rs
Carl-Zama: It reads fine to me, essentially I am calling the private function "multiplication by x" where in this case the x is a secret coefficient or 1.

IceTDrinker: could use some quotes in the comments then

concrete-npe/src/operators.rs
Carl-Zama: I assumed this issue: https://github.com/zama-ai/concrete-core-internal/issues/316 required the full noise formula for the WoP-PBS.

Further, this noise formula does include the fact that bit extraction is performed first however due to the circuit bootstrapping step the noise in the output of the WoP-PBS doesn't depend on the noise during/after bit extraction.

IceTDrinker: Ok, we can probably keep that function for now as it was done, we’ll see if we end up using it in the longer term

concrete-npe/src/operators.rs
IceTDrinker: All right then 🙂

Carl-Zama:
@slab-ci cpu_test

concrete-npe/src/operators.rs
IceTDrinker: @Carl-Zama anything on that ?

IceTDrinker (approved):
Code/doc strings look ok, I let you guys check the maths @samcrx and @Carl-Zama 🙂

concrete-npe/src/operators.rs
Carl-Zama: Indeed as you say, we only use 1 of -1 as a coefficient for pfksk for a circuit bootstrapping so unless this changes in the future the Lipschitz bound on the pfksk will always be 1, hence I chose not to have it as a input, but this could be changed if you prefer?

IceTDrinker: No if this is circuit bootstrap specific it is perfectly fine for me 🙂

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,297 @@ where
...
+///     number_of_ggsw,
+///     poly_size,
+///     glwe_mask_size,
+///     base_log,
+///     level,
+///     dispersion_ggsw,
+///     64,
+/// );
+/// ```
+pub fn estimate_vertical_packing_noise<D>(
```
samcrx, line 1057: Is this the formula for the vertical_packing or for the hybrid_packing ? 

From what I see in the WoP-PBS formula, this must be the hybrid_packing (= vertical packing + final blind rotate). The formula is the same :)

samcrx (commented):
Additional remark: for the doc test if you want to put secure/working parameters, you can use those in https://eprint.iacr.org/2022/704.pdf on page 51. If not, maybe put a comment saying the one in the doctest are just for test purpose but are not correct/secure :)

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,297 @@ where
...
+///     base_log_ksk,
+///     level_ksk,
+///     base_log_bsk,
+///     level_bsk,
+///     64,
+/// );
+/// ```
+#[allow(clippy::too_many_arguments)]
+pub fn estimate_bit_extraction_noise<D1, D2, D3, K1, K2>(
+    number_of_bits_to_extract: ExtractedBitsCount,
```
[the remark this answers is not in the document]
IceTDrinker: I'm curious about why it would be wrong depending on the padding bit or not being in the message ?

concrete-npe/src/operators.rs
Carl-Zama: I was using the function name as given in the `backends/fftw/private/crypto/wop_pbs/mod.rs` file on the am/feature/vertical-packing branch of concrete-core-internal. But by your definition yes, this is hybrid packing, (I admit I did not know the difference between vertical and hybrid packing)

Loris-B: https://eprint.iacr.org/2018/421.pdf
page 27

IceTDrinker: vertical packing it is then thanks @Loris-B 🙏

concrete-npe/src/operators.rs
```diff
@@ -856,6 +858,297 @@ where
...
+        }
+    }
+    output_dispersions.reverse();
+    output_dispersions
+}
+
+/// Computes the dispersion of a circuit bootstrapping for a binary message and a binary secret
+/// key, each ciphertext output is the result of a PBS followed by a private functional
+/// keyswitch where the private function is "multiplication by x" where `x` is a secret key
+/// coefficient (or 1).
```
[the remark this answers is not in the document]
Carl-Zama: Thanks for clarifying this.

concrete-npe/src/operators.rs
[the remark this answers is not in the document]
Carl-Zama: Yeah I am not sure since we extract from LSB to MSB so if there was a bit of padding it would be extracted after the MSB, hence we would not reach it with the correct value of `number_of_bits_to_extract`?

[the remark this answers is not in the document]
samcrx: You're totally right, the padding bit should not change the output of this function. Let me try to give an example:

Let a ciphertext defined like that: 0|m4|m3|m2|m1|0|error. We want to extract the noise after the noise extract of the 2 LSBs. If this function does not know that there will be 2 bits of message + 1 of padding left after the bit extract, this function cannot output the right noise (the shift will not be computed correctly).

The only way to have the correct noise after each bit extract with the current signature of the function is to call it with `number_of_bits_to_extract = 4 +1`. So this function is working only when number_of_bits_to_extract = total number of bit of the message = precision + padding. 

I think we need something more flexible i.e. be able to only compute the noise after 1 bit extract if we only need that

IceTDrinker:
@Carl-Zama this will need a rebase on main we have a refactor on where we put the parameters from concrete-commons 🙂 

Also if you can check with Sam everything is good on the new formula that would be great 🙏

agnesLeroy:
@Carl-Zama would you like me to do the rebase?

Carl-Zama:
> @Carl-Zama would you like me to do the rebase?

@agnesLeroy If you could that would be great, I'm a bit busy this week

agnesLeroy:
> > @Carl-Zama would you like me to do the rebase?
> 
> @agnesLeroy If you could that would be great, I'm a bit busy this week

Alright no worries: have you been able to check with Sam that the formulas looked good to him?

IceTDrinker:
@slab-ci cpu_test

samcrx (approved):

PR #149, opened 2022-08-31
fix(cuda): fix multi-GPU implementation when there are few inputs

agnesLeroy:
### Description
+ Fix multi GPU when there are few inputs: in case the number of GPUs is lower than the number of inputs, restrict the number of GPUs used to the number of inputs.
+ Also, change some types so that it feels more natural to use them.
### Checklist 

### DIFF
--- a/concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_bootstrap_key_conversion.rs
+++ b/concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_bootstrap_key_conversion.rs
@@ -3,8 +3,9 @@ use crate::backends::cuda::implementation::engines::CudaEngine;
 use crate::backends::cuda::implementation::entities::{
     CudaFourierLweBootstrapKey32, CudaFourierLweBootstrapKey64,
 };
-use crate::backends::cuda::private::crypto::bootstrap::CudaBootstrapKey;
-use crate::commons::math::tensor::{AsRefSlice, AsRefTensor};
+use crate::backends::cuda::private::crypto::bootstrap::{
+    convert_lwe_bootstrap_key_from_cpu_to_gpu, CudaBootstrapKey,
+};
 use crate::prelude::{LweBootstrapKey32, LweBootstrapKey64};
 use crate::specification::engines::{
     LweBootstrapKeyConversionEngine, LweBootstrapKeyConversionError,
@@ -72,13 +73,13 @@ impl LweBootstrapKeyConversionEngine<LweBootstrapKey32, CudaFourierLweBootstrapK
         check_poly_size!(poly_size);
         let glwe_dim = input.0.glwe_size().to_glwe_dimension();
         check_glwe_dim!(glwe_dim);
+        let data_per_gpu = input.glwe_dimension().to_glwe_size().0
+            * input.glwe_dimension().to_glwe_size().0
+            * input.input_lwe_dimension().0
+            * input.decomposition_level_count().0
+            * input.polynomial_size().0;
+        let size = data_per_gpu as u64 * std::mem::size_of::<u32>() as u64;
         for stream in self.streams.iter() {
-            let data_per_gpu = input.glwe_dimension().to_glwe_size().0
-                * input.glwe_dimension().to_glwe_size().0
-                * input.input_lwe_dimension().0
-                * input.decomposition_level_count().0
-                * input.polynomial_size().0;
-            let size = data_per_gpu as u64 * std::mem::size_of::<u32>() as u64;
             stream.check_device_memory(size)?;
         }
         Ok(unsafe { self.convert_lwe_bootstrap_key_unchecked(input) })
@@ -88,31 +89,11 @@ impl LweBootstrapKeyConversionEngine<LweBootstrapKey32, CudaFourierLweBootstrapK
         &mut self,
         input: &LweBootstrapKey32,
     ) -> CudaFourierLweBootstrapKey32 {
-        // Copy the entire input vector over all GPUs
-        let mut vecs = Vec::with_capacity(self.get_number_of_gpus() as usize);
-        // TODO
-        //   Check if it would be better to have GPU 0 compute the BSK and copy it back to the
-        //   CPU, then copy the BSK to the other GPUs. Are we sure the BSK generated on each GPU
-        //   will be exactly the same?
-        let total_polynomials = input.input_lwe_dimension().0
-            * input.glwe_dimension().to_glwe_size().0
-            * input.glwe_dimension().to_glwe_size().0
-            * input.decomposition_level_count().0;
-        let alloc_size = total_polynomials * input.polynomial_size().0;
-        for stream in self.streams.iter() {
-            stream.initialize_twiddles(input.polynomial_size());
-            let mut d_vec = stream.malloc::<f64>(alloc_size as u32);
-            let input_slice = input.0.as_tensor().as_slice();
-            stream.convert_lwe_bootstrap_key_32(
-                &mut d_vec,
-                input_slice,
-                input.input_lwe_dimension(),
-                input.glwe_dimension(),
-                input.decomposition_level_count(),
-                input.polynomial_size(),
-            );
-            vecs.push(d_vec);
-        }
+        let vecs = convert_lwe_bootstrap_key_from_cpu_to_gpu::<u32, _>(
+            self.get_cuda_streams(),
+            &input.0,
+            self.get_number_of_gpus(),
+        );
         CudaFourierLweBootstrapKey32(CudaBootstrapKey::<u32> {
             d_vecs: vecs,
             polynomial_size: input.polynomial_size(),
@@ -179,13 +160,13 @@ impl LweBootstrapKeyConversionEngine<LweBootstrapKey64, CudaFourierLweBootstrapK
         check_poly_size!(poly_size);
         let glwe_dim = input.0.glwe_size().to_glwe_dimension();
         check_glwe_dim!(glwe_dim);
+        let data_per_gpu = input.glwe_dimension().to_glwe_size().0
+            * input.glwe_dimension().to_glwe_size().0
+            * input.input_lwe_dimension().0
+            * input.decomposition_level_count().0
+            * input.polynomial_size().0;
+        let size = data_per_gpu as u64 * std::mem::size_of::<u64>() as u64;
         for stream in self.streams.iter() {
-            let data_per_gpu = input.glwe_dimension().to_glwe_size().0
-                * input.glwe_dimension().to_glwe_size().0
-                * input.input_lwe_dimension().0
-                * input.decomposition_level_count().0
-                * input.polynomial_size().0;
-            let size = data_per_gpu as u64 * std::mem::size_of::<u64>() as u64;
             stream.check_device_memory(size)?;
         }
         Ok(unsafe { self.convert_lwe_bootstrap_key_unchecked(input) })
@@ -195,31 +176,11 @@ impl LweBootstrapKeyConversionEngine<LweBootstrapKey64, CudaFourierLweBootstrapK
         &mut self,
         input: &LweBootstrapKey64,
     ) -> CudaFourierLweBootstrapKey64 {
-        // Copy the entire input vector over all GPUs
-        let mut vecs = Vec::with_capacity(self.get_number_of_gpus() as usize);
-        // TODO
-        //   Check if it would be better to have GPU 0 compute the BSK and copy it back to the
-        //   CPU, then copy the BSK to the other GPUs. Are we sure the BSK generated on each GPU
-        //   will be exactly the same?
-        let total_polynomials = input.input_lwe_dimension().0
-            * input.glwe_dimension().to_glwe_size().0
-            * input.glwe_dimension().to_glwe_size().0
-            * input.decomposition_level_count().0;
-        let alloc_size = total_polynomials * input.polynomial_size().0;
-        for stream in self.streams.iter() {
-            stream.initialize_twiddles(input.polynomial_size());
-            let mut d_vec = stream.malloc::<f64>(alloc_size as u32);
-            let input_slice = input.0.as_tensor().as_slice();
-            stream.convert_lwe_bootstrap_key_64(
-                &mut d_vec,
-                input_slice,
-                input.input_lwe_dimension(),
-                input.glwe_dimension(),
-                input.decomposition_level_count(),
-                input.polynomial_size(),
-            );
-            vecs.push(d_vec);
-        }
+        let vecs = convert_lwe_bootstrap_key_from_cpu_to_gpu::<u64, _>(
+            self.get_cuda_streams(),
+            &input.0,
+            self.get_number_of_gpus(),
+        );
         CudaFourierLweBootstrapKey64(CudaBootstrapKey::<u64> {
             d_vecs: vecs,
             polynomial_size: input.polynomial_size(),
--- a/concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_ciphertext_vector_conversion.rs
+++ b/concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_ciphertext_vector_conversion.rs
@@ -2,10 +2,14 @@ use crate::backends::cuda::implementation::engines::{CudaEngine, CudaError};
 use crate::backends::cuda::implementation::entities::{
     CudaLweCiphertextVector32, CudaLweCiphertextVector64,
 };
-use crate::backends::cuda::private::crypto::lwe::list::CudaLweList;
+use crate::backends::cuda::private::crypto::lwe::list::{
+    copy_lwe_ciphertext_vector_from_cpu_to_gpu, copy_lwe_ciphertext_vector_from_gpu_to_cpu,
+    CudaLweList,
+};
+use crate::backends::cuda::private::device::GpuIndex;
+use crate::backends::cuda::private::{compute_number_of_samples_on_gpu, number_of_active_gpus};
 use crate::commons::crypto::lwe::LweList;
-use crate::commons::math::tensor::{AsRefSlice, AsRefTensor};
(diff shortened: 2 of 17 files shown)


agnesLeroy:
@slab-ci gpu_test

agnesLeroy:
@slab-ci cpu_test

agnesLeroy:
@slab-ci cpu_test

concrete-core/src/backends/cuda/implementation/engines/mod.rs
```diff
@@ -71,3 +73,18 @@ impl From<CudaError> for LweCiphertextVectorDiscardingBootstrapError<CudaError>
...
+    number_of_gpus: NumberOfGpus,
+    lwe_ciphertext_count: LweCiphertextCount,
+) -> (NumberOfSamples, NumberOfGpus) {
+    // In case there are less GPUs than inputs, use as many GPUs as there are inputs
+    let mut n = number_of_gpus.0;
+    if n < lwe_ciphertext_count.0 {
+        n = min(lwe_ciphertext_count.0, number_of_gpus.0)
+    }
+    (
+        NumberOfSamples(lwe_ciphertext_count.0 / n as usize),
```
IceTDrinker, line 87: looks dangerours as lwe ct count may not be divisible by n so there will be a remainder ?

Not sure if this function is supposed to manage it or not

concrete-core/src/backends/cuda/implementation/engines/mod.rs
```diff
@@ -71,3 +73,18 @@ impl From<CudaError> for LweCiphertextVectorDiscardingBootstrapError<CudaError>
...
+
+pub(crate) fn compute_number_of_samples_on_all_gpus(
+    number_of_gpus: NumberOfGpus,
+    lwe_ciphertext_count: LweCiphertextCount,
+) -> (NumberOfSamples, NumberOfGpus) {
+    // In case there are less GPUs than inputs, use as many GPUs as there are inputs
+    let mut n = number_of_gpus.0;
+    if n < lwe_ciphertext_count.0 {
+        n = min(lwe_ciphertext_count.0, number_of_gpus.0)
+    }
```
IceTDrinker, line 85: This does not look straight forward to read to me, I think a big if else would be more readable though I think I see the logic

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/mod.rs
```diff
@@ -61,19 +60,22 @@ impl CudaEngine {
...
+        samples_per_gpu: NumberOfSamples,
+        lwe_ciphertext_count: LweCiphertextCount,
+        gpu_index: GpuIndex,
+    ) -> NumberOfSamples {
+        let mut samples = samples_per_gpu.0;
+        if gpu_index.0 > lwe_ciphertext_count.0 {
+            samples = 0;
+        }
+        if gpu_index.0 == self.get_number_of_gpus().0 - 1
+            && lwe_ciphertext_count.0 % self.get_number_of_gpus().0 != 0
```
IceTDrinker, line 74: I'm a bit confused with the existence of the other function then 🤔

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_ciphertext_vector_discarding_bootstrap.rs
```diff
@@ -138,15 +138,17 @@ impl
...
+        );
 
-        for gpu_index in 0..self.get_number_of_gpus() {
-            let mut samples = samples_per_gpu;
-            if gpu_index == self.get_number_of_gpus() - 1
-                && input.lwe_ciphertext_count().0 % self.get_number_of_gpus() as usize != 0
+        for gpu_index in 0..number_of_gpus.0 {
+            let mut samples = samples_per_gpu.0;
+            if gpu_index == number_of_gpus.0 - 1
+                && input.lwe_ciphertext_count().0 % number_of_gpus.0 as usize != 0
```
IceTDrinker, line 149: not sure how it could be written but this looks like some code in an other place to manage the number of samples per GPU taking the remainder into account ?

concrete-core/src/backends/cuda/implementation/engines/mod.rs
agnesLeroy: In case the number of input ciphertexts is lower than the number of GPUs the integer division yields 0, which means that basically there'll be no computation. So in that case I restrict the number of GPUs to the number of inputs (and there'll be one input dealt with per GPU). Now reading it again I realize the condition is written the wrong way round :scream: I'll fix it.

concrete-core/src/backends/cuda/implementation/engines/mod.rs
agnesLeroy: That's normal because only the last GPU takes the remainder: this should be handled correctly (although I never had the opportunity to test it).

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/mod.rs
agnesLeroy: The last GPU takes the remainder of the division: it'll be a bit more loaded than the others.

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_ciphertext_vector_discarding_bootstrap.rs
agnesLeroy: Yes I'm repeating the same code in several places which sould be refactored.

agnesLeroy:
@slab-ci gpu_test

agnesLeroy:
@slab-ci cpu_test

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_ciphertext_vector_discarding_bootstrap.rs
IceTDrinker: ok, could be good to have an issue to not forget !

agnesLeroy: https://github.com/zama-ai/concrete-core-internal/issues/389

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/mod.rs
IceTDrinker: maybe I misread the other function or perhaps a slight rename could help, not sure

concrete-core/src/backends/cuda/implementation/engines/mod.rs
IceTDrinker: ok !

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/mod.rs
agnesLeroy: I'll try to keep only one function, this should simplify the code.

agnesLeroy:
@slab-ci gpu_test

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_bootstrap_key_conversion.rs
```diff
@@ -72,13 +73,13 @@ impl LweBootstrapKeyConversionEngine<LweBootstrapKey32, CudaFourierLweBootstrapK
         check_poly_size!(poly_size);
         let glwe_dim = input.0.glwe_size().to_glwe_dimension();
         check_glwe_dim!(glwe_dim);
+        let data_per_gpu = input.glwe_dimension().to_glwe_size().0
+            * input.glwe_dimension().to_glwe_size().0
+            * input.input_lwe_dimension().0
+            * input.decomposition_level_count().0
+            * input.polynomial_size().0;
+        let size = data_per_gpu as u64 * std::mem::size_of::<u32>() as u64;
```
IceTDrinker, line 81: there already is the need for the compiler in practice https://github.com/zama-ai/concrete-core-internal/issues/188

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
```diff
@@ -22,3 +31,136 @@ pub(crate) struct CudaBootstrapKey<T> {
     // Field to hold type T
     pub(crate) _phantom: PhantomData<T>,
 }
+
+pub(crate) unsafe fn convert_lwe_bootstrap_key_from_cpu_to_gpu<T: UnsignedInteger>(
+    streams: &Vec<CudaStream>,
+    input: &StandardBootstrapKey<Vec<T>>,
```
IceTDrinker, line 37: can probably made generic over the container to avoid headaches with views and whatnot

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
```diff
@@ -22,3 +31,136 @@ pub(crate) struct CudaBootstrapKey<T> {
...
+    number_of_gpus: NumberOfGpus,
+) -> Vec<CudaVec<f64>> {
+    // Copy the entire input vector over all GPUs
+    let mut vecs = Vec::with_capacity(number_of_gpus.0);
+    // TODO
+    //   Check if it would be better to have GPU 0 compute the BSK and copy it back to the
+    //   CPU, then copy the BSK to the other GPUs. Are we sure the BSK generated on each GPU
+    //   will be exactly the same?
```
IceTDrinker, line 45: most likely you will get some relative deviations on the order of 10^-7 for floats and 10^-15 IIRC for doubles as the computation order over several threads can yield slightly different results

IIRC it's a known effect when people use CUDA in ML tasks

concrete-core/src/backends/cuda/private/mod.rs
```diff
@@ -1,4 +1,34 @@
...
+    number_of_gpus: NumberOfGpus,
+    lwe_ciphertext_count: CiphertextCount,
+    gpu_index: GpuIndex,
+) -> NumberOfSamples {
+    let samples_per_gpu = lwe_ciphertext_count.0 / number_of_gpus.0;
+    let mut samples = samples_per_gpu;
+    // We add the remainder of the integer division lwe_count/num_gpus to the load of the last GPU
+    if gpu_index.0 == number_of_gpus.0 - 1 && lwe_ciphertext_count.0 % number_of_gpus.0 != 0 {
+        samples += lwe_ciphertext_count.0 - samples_per_gpu * number_of_gpus.0;
+    }
```
IceTDrinker, line 32: maybe :
```rust
if gpu_index.0 == number_of_gpus.0 - 1 {
        samples += lwe_ciphertext_count.0 % number_of_gpus.0;
}
```

concrete-core/src/backends/cuda/private/device.rs
```diff
@@ -331,81 +275,88 @@ impl CudaStream {
...
+                lwe_dimension.0 as u32,
+                polynomial_size.0 as u32,
+                base_log.0 as u32,
+                level.0 as u32,
+                num_samples.0 as u32,
+                num_samples.0 as u32,
+                lwe_idx.0 as u32,
+                max_shared_memory.0 as u32,
+            )
+        } else if T::BITS == 64 {
```
IceTDrinker, line 295: else if or else ? you do something sliglthly different below, it looks inconsistent between different functions structured in the same way

concrete-core/src/backends/cuda/private/crypto/lwe/list.rs
```diff
@@ -27,3 +32,59 @@ pub(crate) struct CudaLweList<T: UnsignedInteger> {
...
+
+    let mut output =
+        vec![T::ZERO; input.lwe_dimension.to_lwe_size().0 * input.lwe_ciphertext_count.0];
+    for (gpu_index, chunks) in output.chunks_exact_mut(data_per_gpu).enumerate() {
+        let stream = &streams[gpu_index];
+        stream.copy_to_cpu::<T>(chunks, input.d_vecs.get(gpu_index).unwrap());
+    }
+    let last_chunk = output.chunks_exact_mut(data_per_gpu).into_remainder();
+    let last_stream = streams.last().unwrap();
+    last_stream.copy_to_cpu::<T>(last_chunk, input.d_vecs.last().unwrap());
```
IceTDrinker, line 88: sorry to be annoying, does copy_to_cpu manage the case where the remainder might be empty ?

IceTDrinker (commented):
Generally looks good 🙂 great effort !

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
agnesLeroy: Yes, it's just I don't know whether this has a visible effect on the noise after the bootstrap, so I've left the comment to not forget :thinking:

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
agnesLeroy: Done locally

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
agnesLeroy: Actually if we're using the exact same input on the same model of GPU with the same compiled code, the order in which threads are executed should be deterministic. So we should only see differences when GPUs are not the same I think. @pdroalves would you agree?

IceTDrinker: I'm not sure dure to frequency changes in the chip, basically as long as it's not a sequential execution on a monothreaded CPU I would expect deviations in float outputs

concrete-core/src/backends/cuda/private/device.rs
agnesLeroy: I've left the else if everywhere (done locally)

concrete-core/src/backends/cuda/private/crypto/lwe/list.rs
agnesLeroy: You're right it doesn't! I need to add a check here.

concrete-core/src/backends/cuda/private/crypto/bootstrap/mod.rs
agnesLeroy: Edit: actually @IceTDrinker you're right: the warp scheduling is not deterministic: https://stackoverflow.com/questions/24977294/is-cuda-warp-scheduling-deterministic.

agnesLeroy:
@slab-ci gpu_test

agnesLeroy:
@slab-ci gpu_test

IceTDrinker (commented):
A few things still (sorry)

concrete-core/src/backends/cuda/private/crypto/lwe/list.rs
```diff
@@ -27,3 +32,64 @@ pub(crate) struct CudaLweList<T: UnsignedInteger> {
...
+    );
+    let data_per_gpu = samples_per_gpu.0 * input.lwe_dimension.to_lwe_size().0;
+
+    let mut output =
+        vec![T::ZERO; input.lwe_dimension.to_lwe_size().0 * input.lwe_ciphertext_count.0];
+    for (gpu_index, chunks) in output.chunks_exact_mut(data_per_gpu).enumerate() {
+        let stream = &streams[gpu_index];
+        stream.copy_to_cpu::<T>(chunks, input.d_vecs.get(gpu_index).unwrap());
+    }
+    if samples_per_gpu.0 < input.lwe_ciphertext_count.0 {
```
IceTDrinker, line 89: this would always be the case unless lwe_ciphertext_count == 1 ? 🤔

concrete-core/src/backends/cuda/implementation/engines/cuda_engine/lwe_ciphertext_vector_conversion.rs
```diff
@@ -66,15 +70,19 @@ impl LweCiphertextVectorConversionEngine<LweCiphertextVector32, CudaLweCiphertex
...
+            let stream = &self.streams[gpu_index];
+            let samples = compute_number_of_samples_on_gpu(
+                self.get_number_of_gpus(),
+                CiphertextCount(input.lwe_ciphertext_count().0),
+                GpuIndex(gpu_index),
             );
-            let data_per_gpu = samples * input.lwe_dimension().to_lwe_size().0;
-            let size = data_per_gpu as u64 * std::mem::size_of::<u32>() as u64;
+            let data_per_gpu = samples.0 * input.lwe_dimension().to_lwe_size().0;
+            let size = data_per_gpu as u64 * std::mem::size_of::<u64>() as u64;
```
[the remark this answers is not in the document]
agnesLeroy: Nice catch, I checked the other ones they look good

concrete-core/src/backends/cuda/private/crypto/lwe/list.rs
IceTDrinker: or n_gpu >1

agnesLeroy: Totally right, I can replace with:
```
if samples_per_gpu.0 * number_of_available_gpus.0 < input.lwe_ciphertext_count.0
```
instead

IceTDrinker: It’s to check the remainder is non zero ? 

Then yes that should do the trick 🙂

agnesLeroy:
@slab-ci gpu_test

agnesLeroy:
@IceTDrinker normally I took all the comments into account, GPU tests are green! :slightly_smiling_face:

IceTDrinker (approved):
Looks good as far as I can tell !

For the twiddle thing can it be something that the fft does by itself ?

agnesLeroy:
> Looks good as far as I can tell !
> 
> For the twiddle thing can it be something that the fft does by itself ?

Maybe, we need to do some benchmarking to check the effect on performance.

PR #151, opened 2022-09-01
feat(concrete_core_representation): add concrete-core sources analysis

aPere3:
### Resolves:

closes zama-ai/concrete-core-internal#385

### Description

This commit introduces a new crate `concrete-core-representation` that contains tools to analyze the `concrete-core` sources, and build a _Concrete Core Represetation_ out of it. This representation is a tree that builds on top of a `syn` ast with specific nodes that better describes the architecture of the `concrete-core` sources. This representation can then be used for various tasks, among which, automatically generating apis.

### Checklist 

### DIFF
--- a/concrete-core-representation/bin/dump.rs
+++ b/concrete-core-representation/bin/dump.rs
@@ -1,0 +1,10 @@
+use concrete_core_representation::{dump_ccr_to_file, load_ccr};
+
+mod root;
+
+fn main() {
+    let path = std::env::args()
+        .nth(1)
+        .unwrap_or_else(|| String::from("/tmp/ccr_dump.json"));
+    dump_ccr_to_file(path.as_str(), load_ccr(root::get_concrete_core_root()));
+}
--- a/concrete-core-representation/bin/root.rs
+++ b/concrete-core-representation/bin/root.rs
@@ -1,0 +1,11 @@
+use std::path::PathBuf;
+
+pub fn get_concrete_core_root() -> PathBuf {
+    PathBuf::from(file!())
+        .parent()
+        .unwrap()
+        .join("../..")
+        .canonicalize()
+        .unwrap()
+        .join("concrete-core/src/lib.rs")
+}
--- a/concrete-core-representation/src/ccr/abstract_engine_trait_impl.rs
+++ b/concrete-core-representation/src/ccr/abstract_engine_trait_impl.rs
@@ -1,0 +1,111 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing an implementation of the `AbstractEngine` trait.
+#[derive(Serialize, Clone, Debug)]
+pub struct AbstractEngineTraitImpl {
+    pub cfg: CfgStack,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub item_impl: syn::ItemImpl,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub engine_type_ident: syn::Ident,
+}
+
+impl AbstractEngineTraitImpl {
+    /// From an `ItemMod` syn ast node, pointing to a module in a backend, recursively extracts all
+    /// the `AbstractEngineTraitImpl`.
+    pub fn extract_all(
+        cfg_stack_so_far: &CfgStack,
+        module: &syn::ItemMod,
+    ) -> Vec<AbstractEngineTraitImpl> {
+        // INVARIANT: engine types are identified with their original identifier in the abstract
+        // engine impl, not with a path, nor an alias.
+        // INVARIANT: the abstract engine trait is identified by its original identifier
+        // `AbstractEngine`
+        let mut output = Vec::new();
+        if module.content.is_none() {
+            return output;
+        }
+
+        for item in module.content.as_ref().unwrap().1.iter() {
+            // If the item is a module, we recurse:
+            if let syn::Item::Mod(item_mod) = item {
+                output.extend(
+                    AbstractEngineTraitImpl::extract_all(
+                        &cfg_stack_so_far.concat_attrs(item_mod.attrs.as_slice()),
+                        item_mod,
+                    )
+                    .into_iter(),
+                );
+                continue;
+            }
+
+            // If the item is an abstract engine impl, we add it.
+            let maybe_item_impl = probe!(
+                Some(item),
+                syn::Item::Impl(item_impl) => item_impl
+            );
+            let maybe_abstract_engine_impl = probe!(
+                maybe_item_impl,
+                item_impl >> item_impl.trait_.as_ref(),
+                trait_ >> trait_.1.segments.first(),
+                trait_path -> &trait_path.ident,
+                trait_ident ?> *trait_ident == "AbstractEngine",
+                X> maybe_item_impl,
+                item_impl -> item_impl.self_ty.as_ref(),
+                syn::Type::Path(self_path) => self_path,
+                self_path ?> self_path.path.segments.len() == 1,
+                self_path >> self_path.path.segments.first(),
+                self_segment -> &self_segment.ident,
+                self_ident >> pack_somes!(Some(self_ident), maybe_item_impl),
+                tuple -> {
+                    let (engine_type_ident, item_impl) = tuple;
+                    AbstractEngineTraitImpl{
+                        cfg: cfg_stack_so_far.concat_attrs(item_impl.attrs.as_slice()),
+                        item_impl: item_impl.to_owned(),
+                        engine_type_ident: engine_type_ident.to_owned()
+                    }
+                }
+            );
+            if let Some(abstract_engine_trait_impl) = maybe_abstract_engine_impl {
+                output.push(abstract_engine_trait_impl);
+            }
+        }
+        output
+    }
+
+    /// Returns the `syn` node pointing to the constructor of the implementation.
+    pub fn get_constructor(&self) -> &syn::ImplItemMethod {
+        // INVARIANT: The constructor is called `new`
+        self.item_impl
+            .items
+            .iter()
+            .find_map(|item| {
+                probe!(
+                    Some(item),
+                    syn::ImplItem::Method(impl_item_method) => impl_item_method,
+                    impl_item_method ?> impl_item_method.sig.ident == "new"
+                )
+            })
+            .unwrap()
+    }
+
+    /// Returns the `syn` node pointing to the `Parameters` associated type of the implementation.
+    pub fn get_parameters_associated_type(&self) -> &syn::Type {
+        // INVARIANT: The constructor take a single parameter whose type is the associated
+        // `Parameters` type.
+        // INVARIANT: The `Parameters` type is either a Path or a tuple of paths.
+        self.item_impl
+            .items
+            .iter()
+            .find_map(|item| {
+                probe!(
+                    Some(item),
+                    syn::ImplItem::Type(impl_item_type) => impl_item_type,
+                    impl_item_type ?> impl_item_type.ident == "Parameters",
+                    impl_item_type -> &impl_item_type.ty
+                )
+            })
+            .unwrap()
+    }
+}
--- a/concrete-core-representation/src/ccr/abstract_entity_trait_impl.rs
+++ b/concrete-core-representation/src/ccr/abstract_entity_trait_impl.rs
@@ -1,0 +1,74 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing an implementation of the `AbstractEntity` trait.
+#[derive(Serialize, Clone, Debug)]
+pub struct AbstractEntityTraitImpl {
+    pub cfg: CfgStack,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub item_impl: syn::ItemImpl,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub entity_type_ident: syn::Ident,
+}
+
+impl AbstractEntityTraitImpl {
+    /// From an `ItemMod` syn ast node, pointing to a module in a backend, recursively extracts all
+    /// the `AbstractEntityTraitImpl`.
+    pub(crate) fn extract_all(
+        cfg_stack_so_far: &CfgStack,
+        module: &syn::ItemMod,
+    ) -> Vec<AbstractEntityTraitImpl> {
+        // INVARIANT: entity types are identified with their original identifier in the abstract
+        // entity impl, not with a path, nor an alias.
+        // INVARIANT: the abstract entity trait is identified by its original identifier
+        // `AbstractEntity`
+
+        let mut output = Vec::new();
+        if module.content.is_none() {
+            return output;
+        }
+
+        for item in module.content.as_ref().unwrap().1.iter() {
+            // If the item is a module, we recurse.
+            if let syn::Item::Mod(item_mod) = item {
+                output.extend(AbstractEntityTraitImpl::extract_all(
+                    &cfg_stack_so_far.concat_attrs(item_mod.attrs.as_slice()),
+                    item_mod,
+                ));
+                continue;
+            }
+
+            // If the item is an abstract entity impl, we add it to the output.
+            let maybe_item_impl = probe!(
+                Some(item),
+                syn::Item::Impl(item_impl) => item_impl
+            );
+            let maybe_abstract_entity_impl = probe!(
+                maybe_item_impl,
+                item_impl >> item_impl.trait_.as_ref(),
+                trait_ >> trait_.1.segments.first(),
+                trait_path -> &trait_path.ident,
+                trait_ident ?> *trait_ident == "AbstractEntity",
+                X> maybe_item_impl,
+                item_impl -> item_impl.self_ty.as_ref(),
+                syn::Type::Path(self_path) => self_path,
+                self_path ?> self_path.path.segments.len() == 1,
+                self_path >> self_path.path.segments.first(),
+                self_segment -> &self_segment.ident,
+                self_ident >> pack_somes!(Some(self_ident), maybe_item_impl),
+                tuple -> {
+                    let (entity_type_ident, item_impl) = tuple;
+                    AbstractEntityTraitImpl{
+                        cfg: cfg_stack_so_far.concat_attrs(item_impl.attrs.as_slice()),
+                        item_impl: item_impl.to_owned(),
+                        entity_type_ident: entity_type_ident.to_owned()
+                    }
+                }
+            );
+            if let Some(node) = maybe_abstract_entity_impl {
+                output.push(node);
+            }
+        }
+        output
+    }
+}
--- a/concrete-core-representation/src/ccr/cfg_stack.rs
+++ b/concrete-core-representation/src/ccr/cfg_stack.rs
@@ -1,0 +1,46 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing a stack of `cfg` attributes.
+///
+/// As we go deeper in the nesting of modules and items, we apply more and more cfg attributes. This
+/// structure represents that.
+#[derive(Serialize, Clone, Debug)]
+pub struct CfgStack(#[serde(serialize_with = "serialize_vec_with_token_string")] Vec<TokenStream2>);
+
+impl CfgStack {
+    /// Creates an empty stack.
+    pub fn empty() -> CfgStack {
+        CfgStack(Vec::new())
+    }
+
+    /// Creates a stack from a slice of syn attributes.
+    pub fn from_attr(attrs: &[syn::Attribute]) -> Self {
+        let mut stack = Self::empty();
+        stack.push_attrs(attrs);
+        stack
+    }
+
+    /// Adds a set of syn attributes to an existing stack.
+    pub fn push_attrs(&mut self, attrs: &[syn::Attribute]) {
+        attrs
+            .iter()
+            .filter_map(|att| {
+                probe!(
+                    Some(att),
+                    att >> att.path.get_ident(),
+                    ident ?> *ident == "cfg",
+                    X> Some(att.tokens.clone())
+                )
+            })
+            .for_each(|ts| self.0.push(ts));
+    }
+
+    /// Creates a new stack which is a copy of the current stack, to which the `attrs` attributes
+    /// were pushed.
+    pub fn concat_attrs(&self, attrs: &[syn::Attribute]) -> CfgStack {
+        let mut output = self.clone();
+        output.push_attrs(attrs);
+        output
+    }
+}
--- a/concrete-core-representation/src/ccr/engine_trait_impl.rs
+++ b/concrete-core-representation/src/ccr/engine_trait_impl.rs
@@ -1,0 +1,132 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing an implementation of an `*Engine` trait.
+#[derive(Serialize, Clone, Debug)]
+pub struct EngineTraitImpl {
+    pub cfg: CfgStack,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub engine_trait_ident: syn::Ident,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub engine_type_ident: syn::Ident,
+    pub checked_method: EngineTraitImplCheckedMethod,
+    pub unchecked_method: EngineTraitImplUncheckedMethod,
+    pub(crate) engine_trait_parameters:
+        ReadyOrNot<Vec<EngineTraitImplGenericArgument>, syn::AngleBracketedGenericArguments>,
+}
+
+impl EngineTraitImpl {
+    /// Return the generic parameters of the engine trait.
+    pub fn engine_trait_parameters(&self) -> &[EngineTraitImplGenericArgument] {
+        self.engine_trait_parameters.as_ref().unwrap().as_slice()
+    }
+}
+
+impl EngineTraitImpl {
+    /// From an `ItemMod` syn ast node, pointing to a module in a backend, recursively extracts all
+    /// the `EngineTraitImpl`.
+    pub fn extract_all(cfg_stack_so_far: &CfgStack, module: &syn::ItemMod) -> Vec<EngineTraitImpl> {
+        // INVARIANT: engine types are identified with their original identifier in the engine impl,
+        // not with a path, nor an alias.
+        // INVARIANT: the engine trait is identified by its original identifier containing `Engine`
+        let mut output = Vec::new();
+        if module.content.is_none() {
+            return output;
+        }
+
+        for item in module.content.as_ref().unwrap().1.iter() {
+            // If the item is a module, we recurse:
+            if let syn::Item::Mod(item_mod) = item {
+                output.extend(
+                    EngineTraitImpl::extract_all(
+                        &cfg_stack_so_far.concat_attrs(item_mod.attrs.as_slice()),
+                        item_mod,
+                    )
+                    .into_iter(),
+                );
+                continue;
+            }
+
+            // If the item is an engine impl, we add it.
+            let maybe_item_impl = probe!(
+                Some(item),
+                syn::Item::Impl(item_impl) => item_impl
+            );
+            let maybe_engine_impl = probe!(
+                maybe_item_impl,
+                item_impl >> item_impl.trait_.as_ref(),
+                trait_ >> trait_.1.segments.first(),
+                trait_path -> &trait_path.ident,
+                trait_ident ?> *trait_ident != "AbstractEngine",
+                trait_ident ?> *trait_ident != "AbstractEngineSeal",
+                trait_ident ?> trait_ident.to_string().contains("Engine"),
+                X> maybe_item_impl
+            );
+            let maybe_engine_trait_ident = probe!(
+                maybe_engine_impl,
+                item_impl >> item_impl.trait_.as_ref(),
+                trait_ >> trait_.1.segments.last(),
+                trait_segment -> trait_segment.ident.to_owned()
+            );
+            let maybe_engine_type_ident = probe!(
+                maybe_engine_impl,
+                item_impl -> item_impl.self_ty.as_ref(),
+                syn::Type::Path(self_ty_path) => self_ty_path,
+                self_ty_path >> self_ty_path.path.segments.first(),
+                self_segment -> self_segment.ident.to_owned()
+            );
+            let maybe_checked_method = probe!(
+                maybe_engine_impl,
+                item_impl -> EngineTraitImplCheckedMethod::extract(cfg_stack_so_far, item_impl)
+            );
+            let maybe_unchecked_method = probe!(
+                maybe_engine_impl,
+                item_impl -> EngineTraitImplUncheckedMethod::extract(cfg_stack_so_far, item_impl)
+            );
+            // The generic arguments can not be processed right away, because we don't yet
+            // know all the items exported by the backend (hence we can not properly sort
+            // them).
+            let maybe_engine_trait_parameters = probe!(
+                maybe_engine_impl,
+                item_impl >> item_impl.trait_.as_ref(),
+                trait_ >> trait_.1.segments.last(),
+                trait_segment -> &trait_segment.arguments,
+                syn::PathArguments::AngleBracketed(generics) => generics,
+                generics -> ReadyOrNot::Not(generics.to_owned())
+            );
+            let maybe_engine_trait_impl = probe!(
+                pack_somes!(
+                    maybe_engine_impl,
+                    maybe_engine_trait_ident,
+                    maybe_checked_method,
+                    maybe_unchecked_method,
+                    maybe_engine_trait_parameters,
+                    maybe_engine_type_ident
+                ),
+                tuple -> {
+                    let (
+                        engine_impl,
+                        engine_trait_ident,
+                        checked_method,
+                        unchecked_method,
+                        engine_trait_parameters,
+                        engine_type_ident
+                    ) = tuple;
+                    EngineTraitImpl{
+                        cfg: cfg_stack_so_far.concat_attrs(engine_impl.attrs.as_slice()),
+                        engine_trait_ident,
+                        checked_method,
+                        unchecked_method,
+                        engine_trait_parameters,
+                        engine_type_ident
+                    }
+                }
+            );
+            if let Some(engine_trait_impl) = maybe_engine_trait_impl {
+                output.push(engine_trait_impl);
+                continue;
+            }
+        }
+        output
+    }
+}
--- a/concrete-core-representation/src/ccr/engine_trait_impl_generic_argument.rs
+++ b/concrete-core-representation/src/ccr/engine_trait_impl_generic_argument.rs
@@ -1,0 +1,51 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing a generic argument of an implementation of an `*Engine` trait.
+#[derive(Serialize, Clone, Debug)]
+pub enum EngineTraitImplGenericArgument {
+    /// The generic argument is an owned entity
+    #[serde(serialize_with = "serialize_with_token_string")]
+    OwnedEntity(syn::Type),
+    /// The generic argument is a view entity
+    #[serde(serialize_with = "serialize_with_token_string")]
+    ViewEntity(syn::Type),
+    /// The generic argument is a mut view entity
+    #[serde(serialize_with = "serialize_with_token_string")]
+    MutViewEntity(syn::Type),
+    /// The generic argument is a config
+    #[serde(serialize_with = "serialize_with_token_string")]
+    Config(syn::Type),
+    /// The generic argument is a numeric
+    #[serde(serialize_with = "serialize_with_token_string")]
+    Numeric(syn::Type),
+    /// The generic argument is a numeric slice
+    #[serde(serialize_with = "serialize_with_token_string")]
+    NumericSlice(syn::Type),
+    /// The generic argument is a numeric mutable slice
+    #[serde(serialize_with = "serialize_with_token_string")]
+    NumericSliceMut(syn::Type),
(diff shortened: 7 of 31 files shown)


agnesLeroy (commented):
Hello @aPere3! Just before I forget: we should maybe add to the PR template an item stating that you checked either that the new contribution will automatically be integrated to this tool, or that it requires a custom treatment instead. We could point to the README of concrete-core-representation, and summarize in the README what assumptions were made when building this tool, and what would not fit directly with it. What do you think?

concrete-core-representation/src/ccr/engine_type_definition.rs
```diff
@@ -0,0 +1,44 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing the definition of an engine type.
+#[derive(Serialize, Clone, Debug)]
+pub struct EngineTypeDefinition {
+    #[serde(skip)]
+    pub cfg: CfgStack,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub item_struct: syn::ItemStruct,
```
agnesLeroy, line 10: Is the engine type the trait implemented from the specification? Why do you need this type?

concrete-core-representation/src/ccr/misc.rs
```diff
@@ -0,0 +1,387 @@
+use quote::ToTokens;
+use serde::ser::SerializeTupleVariant;
+use serde::Serialize;
+use std::fmt::Debug;
+
+/// A type used as a placeholder for a piece of data that can not be placed just yet.
```
agnesLeroy, line 6: I still don't understand why we need ReadyOrNot -> maybe adding an explanation to this comment would help.

concrete-core-representation/src/ccr/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+    "PolynomialSizeLog",
+    "PolynomialCount",
+    "MonomialDegree",
+    "MonomialIndex",
+    "DecompositionBaseLog",
+    "DecompositionLevelCount",
+    "LutCountLog",
+    "ModulusSwitchOffset",
+    "DeltaLog",
+    "ExtractedBitsCount",
```
agnesLeroy, line 84: A good example of things to point out to contributors is to add new parameters here. Beware that I added another new parameter in the bit extract PR by the way: we'll probably merge it before this one. It's only used in a generic check but still we should make sure to add it.

agnesLeroy (commented):
Hello @aPere3! Thanks a lot for this! I went through the PR rather quickly: it's too hard to read more than 2,000 lines of code :sob:. So you'll see I've asked some questions that probably just show my lack of understanding of the thing, so I can understand it a bit more maybe :sweat_smile:

concrete-core-representation/src/misc.rs
```diff
@@ -0,0 +1,90 @@
...
+use std::io::Write;
+
+/// Dumps the `ccr` to a json in `/tmp/ccr_dump.json`. For debugging purpose.
+pub fn dump_ccr<CC: Borrow<ConcreteCore>>(ccr: CC) {
+    let path = "/tmp/ccr_dump.json";
+    let mut file = File::create(&path).unwrap();
+    let output = serde_json::to_string(ccr.borrow()).unwrap();
+    file.write_all(output.as_bytes()).unwrap();
+    file.flush().unwrap();
+}
```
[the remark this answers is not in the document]
aPere3: Hmm that is interesting ! So as to be able to see between two commit what is detected that was not before ?

concrete-core-representation/src/ccr/entity_type_definition.rs
```diff
@@ -0,0 +1,19 @@
...
+use serde::Serialize;
+
+/// A node representing the definition of an entity type.
+#[derive(Serialize, Clone, Debug)]
+pub struct EntityTypeDefinition {
+    #[serde(skip)]
+    pub cfg: CfgStack,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub item_struct: syn::ItemStruct,
+    pub ownership: EntityOwnership,
```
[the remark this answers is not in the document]
aPere3: Because we don't treat views, mut views, and owned entities the same way in the apis.

concrete-core-representation/bin/root.rs
```diff
@@ -0,0 +1,12 @@
+use std::path::PathBuf;
+
+pub fn get_concrete_core_root() -> PathBuf {
+    let mut concrete_core_root = PathBuf::from(file!()).canonicalize().unwrap();
+    concrete_core_root.pop();
+    concrete_core_root.pop();
+    concrete_core_root.pop();
```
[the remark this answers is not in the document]
aPere3: A reference to [Antoine de Caunes emission on Radio France of course](https://www.radiofrance.fr/franceinter/podcasts/popopop) :rofl: !

More seriously, it is `../../..`. Probaby there is something better to do :laughing:

concrete-core-representation/src/ccr/abstract_engine_trait_impl.rs
```diff
@@ -0,0 +1,113 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing an implementation of the `AbstractEngine` trait.
+#[derive(Serialize, Clone, Debug)]
+pub struct AbstractEngineTraitImpl {
+    #[serde(skip)]
+    pub cfg: CfgStack,
```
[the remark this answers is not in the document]
aPere3: Are you talking about the `CfgStack` thing ? This contains the set of `#[cfg(feature="taratata")]` which apply to the item in question (here the `AbstractEngineTraitImpl`). It is a stack because of the nesting of modules, which basically stack such configurations on top of the one of the enclosing module. 

Is that clearer ?

concrete-core-representation/src/ccr/abstract_entity_trait_impl.rs
```diff
@@ -0,0 +1,73 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing an implementation of the `AbstractEntity` trait.
+#[derive(Serialize, Clone, Debug)]
+pub struct AbstractEntityTraitImpl {
+    #[serde(skip)]
+    pub cfg: CfgStack,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub item_impl: syn::ItemImpl,
```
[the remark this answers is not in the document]
aPere3: If you refer to the `engine_type_ident` field of the `AbstractEngineTraitImpl` node, then this is a good question ! 

I believe this is because I did not need the identifier extracted from the trait impl ? Not sure ! 

But just to make things clear: the identifier of the entity type is still accessible in the `item_impl` field. We just have to properly `probe!` it.

concrete-core-representation/src/ccr/engine_type_definition.rs
aPere3: For me:
+ engine trait = an `*Engine` trait
+ engine trait impl = an `impl *Engine for toto` block
+ engine type = a type that implement some `*Engine` traits.
+ engine type definition = the definition of a type that implements `*Engine` traits.

concrete-core-representation/src/ccr/misc.rs
aPere3: See https://github.com/zama-ai/concrete-core/pull/151/files#r963649422

concrete-core-representation/src/misc.rs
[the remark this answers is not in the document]
agnesLeroy: Yes exactly

concrete-core-representation/bin/root.rs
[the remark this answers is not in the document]
agnesLeroy: I guessed so but it looked funny :laughing:

concrete-core-representation/src/ccr/abstract_engine_trait_impl.rs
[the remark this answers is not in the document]
agnesLeroy: Oh ok, thanks for the clarification!

concrete-core-representation/src/ccr/entity_type_definition.rs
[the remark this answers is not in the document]
agnesLeroy: I'll need a little bit more details here sorry :sweat_smile: how are they treated differently?

concrete-core-representation/src/ccr/abstract_entity_trait_impl.rs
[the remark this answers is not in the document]
agnesLeroy: Hum ok, I just thought it was funny to have the two structures be slightly different from one another. If it's possible to have them look the same that'd probably be good I think.

concrete-core-representation/src/ccr/engine_type_definition.rs
agnesLeroy: Oh alright, so DefaultEngine is an engine type if I get it right?

concrete-core-representation/src/ccr/engine_trait_impl.rs
```diff
@@ -0,0 +1,133 @@
...
+    #[serde(skip)]
+    pub cfg: CfgStack,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub engine_trait_ident: syn::Ident,
+    #[serde(serialize_with = "serialize_with_token_string")]
+    pub engine_type_ident: syn::Ident,
+    pub checked_method: EngineTraitImplCheckedMethod,
+    pub unchecked_method: EngineTraitImplUncheckedMethod,
+    pub(crate) engine_trait_parameters:
+        ReadyOrNot<Vec<EngineTraitImplGenericArgument>, syn::AngleBracketedGenericArguments>,
```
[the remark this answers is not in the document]
agnesLeroy: Alright, thanks for the explanation (again :sweat_smile:) I think it'd be worth having this explanation you just wrote in a comment in the code.

concrete-core-representation/src/ccr/engine_type_definition.rs
agnesLeroy: Is this explanation about vocabulary written in the crate documentation? I think it would be good to have it.

concrete-core-representation/src/misc.rs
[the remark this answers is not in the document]
aPere3: Done ! See zama-ai/concrete-core-internal#403

concrete-core-representation/src/ccr/entity_type_definition.rs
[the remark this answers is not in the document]
aPere3: Two examples:
+ In the wasm api, we don't want to expose views and mut views because it does not work with wasm_bindgen,
+ In the c api, we have to expose a lifetime parameter when we expose a view, but not when we expose a owned entity !

concrete-core-representation/bin/root.rs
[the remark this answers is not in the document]
aPere3: Changed it to something clearer. Don't know why I used this pop and push stuffs :laughing:

concrete-core-representation/src/ccr/abstract_entity_trait_impl.rs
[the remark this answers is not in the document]
aPere3: Did it !

concrete-core-representation/src/ccr/engine_type_definition.rs
aPere3: Add it to the `lib.rs` doc.

concrete-core-representation/src/ccr/mod.rs
aPere3: Okay, I'll leave this one open to ensure we add all the needed parameters before we merge.

concrete-core-representation/src/ccr/engine_trait_impl.rs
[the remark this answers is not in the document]
aPere3: I added this explanation to the documentation of `ReadyOrNot`.

concrete-core-representation/src/ccr/abstract_engine_trait_impl.rs
```diff
@@ -0,0 +1,113 @@
...
+                    AbstractEngineTraitImpl{
+                        cfg: cfg_stack_so_far.concat_attrs(item_impl.attrs.as_slice()),
+                        item_impl: item_impl.to_owned(),
+                        engine_type_ident: engine_type_ident.to_owned()
+                    }
+                }
+            );
+            if let Some(abstract_engine_trait_impl) = maybe_abstract_engine_impl {
+                output.push(abstract_engine_trait_impl);
+                continue;
```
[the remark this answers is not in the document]
aPere3: Yes :laughing:

concrete-core-representation/src/ccr/abstract_entity_trait_impl.rs
```diff
@@ -0,0 +1,73 @@
...
+                X>
+                maybe_item_impl,
+                item_impl -> AbstractEntityTraitImpl {
+                        cfg: cfg_stack_so_far.concat_attrs(item_impl.attrs.as_slice()),
+                        item_impl: item_impl.to_owned(),
+                    }
+            );
+            if let Some(node) = maybe_abstract_entity_impl {
+                output.push(node);
+                continue;
```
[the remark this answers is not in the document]
aPere3: Done.

concrete-core-representation/src/ccr/cfg_stack.rs
```diff
@@ -0,0 +1,55 @@
...
+                .filter_map(|att| {
+                    probe!(
+                        Some(att),
+                        att >> att.path.get_ident(),
+                        ident ?> *ident == "cfg",
+                        X> Some(att.tokens.clone())
+                    )
+                })
+                .collect(),
+        )
```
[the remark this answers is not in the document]
aPere3: Yes clearly !

concrete-core-representation/src/ccr/struct_definition.rs
```diff
@@ -0,0 +1,53 @@
...
+                syn::Item::Struct(item_struct) => item_struct,
+                item_struct ?> matches!(item_struct.vis, syn::Visibility::Public(_)),
+                item_struct -> item_struct
+            );
+            if let Some(item_struct) = maybe_struct {
+                output.push(StructDefinition {
+                    cfg: cfg_stack_so_far.concat_attrs(item_struct.attrs.as_slice()),
+                    item_struct: item_struct.to_owned(),
+                });
+                continue;
```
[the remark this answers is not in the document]
aPere3: Done !

concrete-core-representation/src/ccr/entity_type_definition.rs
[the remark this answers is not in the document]
agnesLeroy: :+1: Thank you! I think this explanation would deserve to be in a comment in the code :slightly_smiling_face:

concrete-core-representation/src/ccr/mod.rs
agnesLeroy: The bit extraction feature has been merged so you can rebase & add the new parameter (CiphertextModulusLog).

concrete-core-representation/src/ccr/misc.rs
```diff
@@ -0,0 +1,408 @@
...
+pub(crate) fn pull_first_match<T, F>(vec: &mut Vec<T>, f: F) -> T
+where
+    F: Fn(&T) -> bool,
+{
+    let index = vec.iter().position(f).unwrap();
+    vec.remove(index)
+}
+
+/// Removes all the items matching a filter from a vec, and returns them.
+pub(crate) fn pull_all_match<T, F>(vec: &mut Vec<T>, f: F) -> Vec<T>
```
IceTDrinker, line 76: pull_all_matches

concrete-core-representation/src/ccr/misc.rs
```diff
@@ -0,0 +1,408 @@
...
+where
+    F: Fn(&T) -> bool,
+{
+    let mut output = Vec::new();
+    for index in (0..vec.len()).rev() {
+        if f(&vec[index]) {
+            output.push(vec.remove(index));
+        }
+    }
```
IceTDrinker, line 85: this can crash right ? as removing something at index means we can have an index past the new end of the resized vector ?

concrete-core-representation/src/ccr/misc.rs
```diff
@@ -0,0 +1,408 @@
...
+    output
+}
+
+pub(crate) fn serialize_with_token_string<T, S>(input: &T, serializer: S) -> Result<S::Ok, S::Error>
```
IceTDrinker, line 89: can you elaborate quickly on the need for the three variants ?

concrete-core-representation/src/ccr/misc.rs
```diff
@@ -0,0 +1,408 @@
...
+/// assert_eq!(None, pack_somes!(Some(1), Some(2), None));
+#[macro_export]
+macro_rules! pack_somes {
+    ($i: expr, $ii:expr) => {
+        if let (Some(i), Some(ii)) = ($i, $ii) {
+            Some((i, ii))
+        } else {
+            None
+        }
+    };
```
IceTDrinker, line 149: No way to express this in variadics ?

Or maybe it's just way simpler this way ?

concrete-core-representation/src/ccr/misc.rs
```diff
@@ -0,0 +1,408 @@
...
+///     Even(usize),
+///     Odd(usize)
+/// }
+/// let output = probe!(
+///     Some(10),
+///     val -> val * 2,                            // Applying a map pipe.
+///     new_val ?> new_val % 2 == 0,               // Applying a filter pipe.
+///     X> Some(IntegerParity::Odd(9))             // Applying a reboot pipe.
+///     IntegerParity::Odd(val) => val,            // Applying the variant pipe.
+///     val >> val.checked_add(55),                // Applying a then pipe.
```
IceTDrinker, line 216: I would replace all symbols by human readable keywords

so like 


map
filt
take/new/restart/yield/use/let
the fat arrow => for variant is ok I think as it's already part of the rust syntax
then

why is the then pipe required compared to the map one ?

concrete-core-representation/src/ccr/inline.rs
```diff
@@ -0,0 +1,66 @@
...
+    read_file_and_inline_modules(path, |path| {
+        let forbidden_files = ["commons", "private"];
+        for f in forbidden_files.iter() {
+            if path.as_os_str().to_str().unwrap().contains(f) {
+                return false;
+            }
+        }
+        true
+    })
+}
```
IceTDrinker, line 18: We may want to have a cfg directive or some stuff in the source itself to know if we want to parse the module in the long term ? So that the source of truth is concrete-core and the core analysis just performs the analysis

concrete-core-representation/src/ccr/inline.rs
```diff
@@ -0,0 +1,66 @@
...
+    file.read_to_string(&mut content).unwrap();
+    let mut ast = syn::parse_file(&content).unwrap();
+
+    for item in ast.items.iter_mut() {
+        if let syn::Item::Mod(ref mut module) = item {
+            if module.content.is_none() {
+                *module = inline_module(path.as_ref(), module, filter);
+            }
+        }
+    }
```
IceTDrinker, line 35: this looks like we may reload a module several times ?

concrete-core-representation/src/ccr/struct_definition.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+impl StructDefinition {
+    /// From a `ItemMod` syn ast node, pointing to a module in a backend, recursively extract all
+    /// the public structure definitions.
+    pub(crate) fn extract_all(
+        cfg_stack_so_far: &CfgStack,
+        module: &syn::ItemMod,
+    ) -> Vec<StructDefinition> {
+        // INVARIANT: concrete-core items are publicly visible.
```
IceTDrinker, line 17: do we check for that invariant ? (could be interesting for some automated PR checks to ease reviews)

concrete-core-representation/src/ccr/entity_trait_impl.rs
```diff
@@ -0,0 +1,68 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing an implementation of an `*Entity` trait.
+#[derive(Serialize, Clone, Debug)]
+pub struct EntityTraitImpl {
+    #[serde(skip)]
+    pub cfg: CfgStack,
```
IceTDrinker, line 8: maybe we want to serialize this in the json we dump ? unsure (applies to all the cfg stack stuff)

concrete-core-representation/src/ccr/entity_trait_impl.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub item_impl: syn::ItemImpl,
+}
+
+impl EntityTraitImpl {
+    /// From an `ItemMod` syn ast node, pointing to a module inside a backend, recursively extract
+    /// all the `EntityTraitImpl` nodes.
+    pub fn extract_all(cfg_stack_so_far: &CfgStack, module: &syn::ItemMod) -> Vec<EntityTraitImpl> {
+        // INVARIANT: entity types are identified with their original identifier in the entity impl,
+        // not with a path, nor an alias.
+        // INVARIANT: the entity trait is identified by its original identifier containing `Entity`
```
IceTDrinker, line 19: Basically same question for all invariants :

- are those expected to be followed in core (then can we check the invariants are respected)
- are those expected to be conserved in the CCR crate ?

concrete-core-representation/src/ccr/engine_trait_impl_unchecked_method.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+impl EngineTraitImplUncheckedMethod {
+    /// From an `ItemImpl` syn ast node, pointing to an engine trait impl block, extracts the
+    /// `EngineTraitImplUncheckedMethod`.
+    pub fn extract(
+        cfg_stack_so_far: &CfgStack,
+        impl_block: &syn::ItemImpl,
+    ) -> EngineTraitImplUncheckedMethod {
+        // INVARIANT: The unhecked entry point is the first unsafe method.
```
IceTDrinker, line 20: typo : unchecked

concrete-core-representation/src/ccr/engine_trait_impl_return.rs
```diff
@@ -0,0 +1,55 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing the return type of an implementation of an `*Engine` trait.
+#[derive(Serialize, Clone, Debug)]
+pub enum EngineTraitImplReturn {
+    /// The return type is an owned entity
```
IceTDrinker, line 7: a small example for each would help I think 🙂

concrete-core-representation/src/ccr/engine_trait_impl_generic_argument.rs
```diff
@@ -0,0 +1,51 @@
+use super::*;
+use serde::Serialize;
+
+/// A node representing a generic argument of an implementation of an `*Engine` trait.
+#[derive(Serialize, Clone, Debug)]
+pub enum EngineTraitImplGenericArgument {
+    /// The generic argument is an owned entity
```
IceTDrinker, line 7: some example could help here as well ?

IceTDrinker (commented):
Thanks @aPere3 for the PR, that's a big amount of work.

A few comments (some are general to some pattern you use, I let you check them).

I'm a bit worried about the rigidity the ccr may add to concrete-core, basically if we want to change some way we structure core we will have to adapt ccr to take that into account.

While generally I don't necessarily see big structural changes in the short term I think it's something to take into consideration.

I think ccr has great potential to enforce some invariants for PR with a checking mechanism (linked to my question about invariants, it feels like there may be two types of invariants from what you wrote).

What is sure is that this will speed up delivery on stabilized designs, but for anything new (that might go out of the current model, maybe we won't have that case but I prefer to be careful) it's less clear to me

concrete-core-representation/src/ccr/mod.rs
aPere3: Well, we could yes. We would just have to parse the second crate and properly integrate it into the final representation. I expected those things to be brought to core pretty fast, which is why I did not bother doing it for now.

IceTDrinker: yep sounds good !

concrete-core-representation/bin/dump.rs
```diff
@@ -0,0 +1,8 @@
+use concrete_core_representation::{dump_ccr, load_ccr};
+
+mod root;
+
+fn main() {
+    dump_ccr(load_ccr(root::get_concrete_core_root()));
+    println!("/tmp/ccr_dump.json")
+}
```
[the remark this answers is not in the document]
aPere3: Done -> If argument given, then it is expected to be a path. If not, we default to `/tmp/ccr_dump.json`. Nothing too fancy for now, since it is not clear what we want to do with that. Is that okay, if we rework when we have a clearer vision ?

concrete-core-representation/src/ccr/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Builds a representation of concrete-core from the path of the root file.
+pub fn load_ccr<P: AsRef<Path>>(path: P) -> ConcreteCore {
+    let root = inline::read_crate(path);
+    let empty_cfg_stack = CfgStack::empty();
+    ConcreteCore::extract(&empty_cfg_stack, &root)
+}
+
+// TODO: Once reintegrated into concrete-core, those parameters should parsed from the sources as
+// well.
```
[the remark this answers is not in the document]
aPere3: Done

concrete-core-representation/src/ccr/mod.rs
```diff
@@ -0,0 +1,91 @@
+//! A module containing a representation for the concrete-core repository.
+use serde::Serialize;
+use std::fmt::Debug;
+use std::path::Path;
+use syn::__private::TokenStream2;
```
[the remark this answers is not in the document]
IceTDrinker: not sure, I just wince when I see an import from a private part of a crate

concrete-core-representation/src/misc.rs
```diff
@@ -0,0 +1,90 @@
+use crate::{Backend, ConcreteCore, Engine, EngineTraitImpl, Entity};
+use quote::ToTokens;
+use std::borrow::Borrow;
+use std::fs::File;
+use std::io::Write;
+
+/// Dumps the `ccr` to a json in `/tmp/ccr_dump.json`. For debugging purpose.
+pub fn dump_ccr<CC: Borrow<ConcreteCore>>(ccr: CC) {
+    let path = "/tmp/ccr_dump.json";
```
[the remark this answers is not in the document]
aPere3: Solved !

concrete-core-representation/src/misc.rs
```diff
@@ -0,0 +1,90 @@
...
+    let backend_title = format!("{} backend", backend.ident);
+    println!("{}{}", indent_str, backend_title);
+    println!("{}{}", indent_str, "-".repeat(backend_title.len()));
+    println!();
+    print_all_engines(indent + 1, &backend.engines);
+    println!();
+    print_all_entities(indent + 1, &backend.entities);
+    println!();
+}
+
```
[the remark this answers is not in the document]
aPere3: Well, those are only used to print a human readable summary in the `summary` binary target. I did not want to think deeply about how we may want to use the `Display` trait for all those structures, so I just left it there :rofl: . We can move it, but maybe it requires a more careful thinking than what is here.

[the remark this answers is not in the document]
IceTDrinker: sure, again if it's not super important not a priority

concrete-core-representation/src/ccr/misc.rs
```diff
@@ -0,0 +1,408 @@
...
+///
+/// For this reason, this classification can only happen in a second time. For the time being
+/// though, we can store the `syn::AngleBracketedGenericArguments` node, which correspond to the
+/// `<Vec<32>, LweCiphertext32>` region, so as to be able to classify the content after.
+#[derive(Serialize, Clone, Debug)]
+pub(crate) enum ReadyOrNot<R, N> {
+    Ready(R),
+    #[serde(skip)]
+    Not(N),
+}
```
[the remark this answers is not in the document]
aPere3: Kind of! But compared to what usually exist for lazy loading, you can not put the loading logic inside a closure at the level of the `ReadyOrNot`, because it requires a reference to the whole representation, which you don't have access to at the moment you create the `ReadyOrNot`. 

Probably this can be made better, but for now, I found this approach to be simple enough to use.

concrete-core-representation/src/ccr/misc.rs
aPere3: Done !

concrete-core-representation/src/ccr/misc.rs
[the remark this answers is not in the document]
IceTDrinker: yep I'm all for simple !

concrete-core-representation/src/ccr/misc.rs
aPere3: It does not because I iterate in reverse :wink: !

IceTDrinker: indeed ! missed the rev

concrete-core-representation/src/ccr/misc.rs
aPere3: We use a different one depending on how long the tuple is to serialize. See:
+ https://github.com/zama-ai/concrete-core/blob/304a3305a1825a29d5524027c38d7e900382c187/concrete-core-representation/src/ccr/engine_trait_impl_arg.rs#L32
+ https://github.com/zama-ai/concrete-core/blob/304a3305a1825a29d5524027c38d7e900382c187/concrete-core-representation/src/ccr/engine_trait_impl_arg.rs#L29

concrete-core-representation/src/ccr/misc.rs
aPere3: I tried to, but I could not find a way to do it at the time :rofl:. Not remember why. Probably it is possible, but I don't know all the arcane secret of macros for now :laughing:

concrete-core-representation/src/ccr/inline.rs
aPere3: That's a very good idea !

concrete-core-representation/src/ccr/inline.rs
aPere3: Why is that ? To me it looks fine, but maybe I missed something ?

concrete-core-representation/src/ccr/struct_definition.rs
aPere3: Nope, it is more something that we rely on to build the representation.

concrete-core-representation/src/ccr/inline.rs
IceTDrinker: actually I may misunderstand how the ast/syn stores modules

concrete-core-representation/src/ccr/entity_trait_impl.rs
aPere3: I think I did not integrate it to avoid cluttering the dump, but well, it's not like we are going to read the dump without tools anyway !

concrete-core-representation/src/ccr/struct_definition.rs
IceTDrinker: is there a way to check it ? because if the invariant we expect is not enforced the ccr generation will kind of fail right and potentially silently as well

concrete-core-representation/src/ccr/entity_trait_impl.rs
aPere3: Done !

concrete-core-representation/src/ccr/entity_trait_impl.rs
aPere3: Those are supposed to be followed in core !

And I would say that generally, if I wrote it there, it is that it could not be verified (there are some stuffs that I verify already).

concrete-core-representation/src/ccr/engine_trait_impl_unchecked_method.rs
aPere3: Done !

concrete-core-representation/src/ccr/engine_trait_impl_return.rs
aPere3: You mean an example describing the kind of node we refer to ? Like adding something like this :
```rust
/// # Example
/// ```rust
/// impl TotoEngine for SomeEngine{
///     fn perform_toto(...) -> Result<u8, Error>
///                                    ^^ This type.
///     {
///         ...
///     }
///     ...
/// }
/// ```
```

concrete-core-representation/src/ccr/engine_trait_impl_generic_argument.rs
aPere3: Let's discuss that for all nodes on the above comment.

concrete-core-representation/bin/dump.rs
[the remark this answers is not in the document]
aPere3: All good then !

concrete-core-representation/src/ccr/entity_trait_impl.rs
IceTDrinker: supposed vs. actually followed is dangerous IMO 😅 

but that's just me being psycho-rigid

concrete-core-representation/src/ccr/engine_trait_impl_return.rs
IceTDrinker: `/// The return type is an owned entity e.g. LweCiphertext32`

maybe that's overkill though

concrete-core-representation/src/misc.rs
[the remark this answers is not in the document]
aPere3: All good then !

concrete-core-representation/src/ccr/struct_definition.rs
aPere3: Yes ! One way to go could be:  for each PR, have a kind of automatic report on new things added on the public api, that uses ccr, with json diffing or something like that. Then the reviewer verifies that what he sees is well represented in the diff.

IceTDrinker: arf... I don't think we should burden reviewers some more :/

concrete-core-representation/src/ccr/entity_trait_impl.rs
aPere3: No I think you raise a very valid point, but it is kind of a chicken and egg problem actually. If you want to check that an entity is publicly available, you need to know that it is an entity ! And the problem is that to know that it is an entity, you rely on the fact that it is publicly available ! 

That being said, for most invariants, breaking them would mean doing something super weird. For this one for instance, it would mean declaring an entity with something like that:
```rust
struct SomeEntity(...);

pub use SomeEntity;
```

aPere3: Oh no its for another comment but anyway you get the message !

IceTDrinker: yes I was thinking that at some point we probably want some pattern like "this looks like an entity" because it's a struct with "Entity" at the end of it and if it's a public struct it most likely is an Entity, maybe this is not realistic, I'm not sure

concrete-core-representation/src/ccr/struct_definition.rs
aPere3: Then we have to believe :pray: ! hahaha :laughing:

IceTDrinker: yeah, maybe in a future evolution then

concrete-core-representation/src/ccr/engine_trait_impl_return.rs
IceTDrinker: fair enough

concrete-core-representation/src/ccr/entity_type_definition.rs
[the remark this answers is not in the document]
aPere3: Done !

concrete-core-representation/src/ccr/entity_trait_impl.rs
aPere3: Well it turns out that entity __types__ do not have `Entity` at the end of their name. See `LweCiphertext32` for instance. This is why I could not use this approach but yeah, otherwise we could do something like that. 

Another way could be to have some tags like:
```rust
#[ccr(entity-def)]
pub struct LweCiphertext32(..);
```
or something in this spirit. The invariant would then be different: Every entity must have the `#[ccr(entity-def)]` attribute above. In this case the free invariant (the one we can not verify automatically) would be simpler to enforce for maintainers, and we could verify other invariants (like pub visibility of entities) automatically.

IceTDrinker: could be interesting to tag things !

aPere3: Okay, I'll open an issue about that then !

aPere3: zama-ai/concrete-core-internal#408

agnesLeroy:
Hello @aPere3! Is this PR good to go?

concrete-core-representation/src/ccr/misc.rs
aPere3: As discussed today, we live it like that for now, and we will potentially rework it next time someone work on the topic.

aPere3:
@agnesLeroy Yep, ready to go !

agnesLeroy (approved):

PR #156, opened 2022-09-01
Feature/automatic concrete core wasm

aPere3:
### Resolves: 

closes zama-ai/concrete-core-internal#386

### Description

This is a followup to #151, which uses the representation from `concrete-core-representation` to automatically generate `concrete-core-wasm`.

### Checklist 


concrete-core-wasm/Cargo.toml
```diff
@@ -10,15 +10,21 @@ documentation = "https://docs.zama.ai/concrete/core-lib/"
...
 readme = "README.md"
 keywords = ["fully", "homomorphic", "encryption", "fhe", "cryptography"]
+build = "build/build.rs"
 
 [lib]
 crate-type = ["cdylib", "rlib"]
 
+#[[bin]]
+#name = "generate"
+#path = "build/main.rs"
```
IceTDrinker, line 20: to be removed ?

concrete-core-wasm/build/build.rs
```diff
@@ -0,0 +1,74 @@
...
+    write_gen_rs(&formatted_binding);
+
+    Ok(())
+}
+
+fn write_gen_rs(binding: &str) {
+    let path = out_dir().join("__gen.rs");
+    let mut file = File::create(&path).unwrap();
+    file.write_all(binding.as_bytes()).unwrap();
+    file.flush().unwrap();
```
IceTDrinker, line 34: is the flush required if the file gets closed at the end of the scope ?

Probably missed something like this in the other ccr PR

concrete-core-wasm/build/build.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+fn get_concrete_core_root() -> PathBuf {
+    let mut concrete_core_root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
+        .canonicalize()
+        .unwrap();
+    concrete_core_root.pop();
+    concrete_core_root.push("concrete-core");
+    concrete_core_root.push("src");
+    concrete_core_root.push("lib.rs");
```
IceTDrinker, line 48: join + canonicalize ?

concrete-core-wasm/build/build.rs
```diff
@@ -0,0 +1,74 @@
...
+fn format_binding(input: String) -> Result<String, String> {
+    let mut rustfmt = Command::new("rustfmt")
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .spawn()
+        .expect("failed to execute child");
+    let mut stdin = rustfmt
+        .stdin
+        .take()
+        .ok_or_else(|| "Failed tottake stdin".to_string())?;
```
IceTDrinker, line 61: typo : to take

concrete-core-wasm/build/build.rs
```diff
@@ -0,0 +1,74 @@
...
+            .write_all(input.as_bytes())
+            .expect("failed to write to stdin");
+    });
+    let output = rustfmt
+        .wait_with_output()
+        .map_err(|e| format!("Failed to gather rustfmt output: {}", e))?;
+    if !output.status.success() {
+        return Err("Failed to format binding.".to_string());
+    }
+    String::from_utf8(output.stdout).map_err(|e| format!("Failed to read rustfmt output: {}", e))
```
IceTDrinker, line 73: can we maybe give the path of the file to rustfmt directly ?

concrete-core-wasm/build/generation.rs
```diff
@@ -0,0 +1,500 @@
...
+            }
+        }
+    }
+}
+
+fn generate_public_constructor_args(parameter_associated_type: &Type) -> TokenStream2 {
+    // INVARIANT: The parameter associated type is either a path or a tuple of paths.
+
+    // If the parameter is a `Box<dyn Seeder>`
+    let test_ast: syn::TypePath = parse_quote!(Box<dyn Seeder>);
```
IceTDrinker, line 183: this looks awefully specific 🤔

concrete-core-wasm/build/prune.rs
```diff
@@ -0,0 +1,79 @@
+//! This module contains a function that prunes the ccr to only keep the elements that we want to
+//! expose in the wasm api.
+use concrete_core_representation::{
+    ConcreteCore, EngineTraitImplArg, EngineTraitImplReturn, EntityOwnership,
+};
+
+const BLACKLISTED_BACKENDS: [&str; 2] = ["cuda", "fftw"];
```
IceTDrinker, line 7: as you pointed out, maybe should be an opt-in (plus opt-in may allow to configure through features which backends we export for e.g. the C api once we get to it and try to follow a similar pattern)

concrete-core-wasm/src/lib.rs
```diff
@@ -1,13 +1,90 @@
...
+        (DecompositionLevelCount, DecompositionLevelCount, usize),
+        (LweDimension, LweDimension, usize),
+        (LweCiphertextCount, LweCiphertextCount, usize),
+        (LweSize, LweSize, usize),
+        (MonomialIndex, MonomialIndex, usize),
+        (GlweDimension, GlweDimension, usize),
+        (GlweSize, GlweSize, usize),
+        (GlweCiphertextCount, GlweCiphertextCount, usize),
+        (PolynomialSize, PolynomialSize, usize)
+    }
```
IceTDrinker, line 88: I don't remember why this needs to be done by hand ?

concrete-core-wasm/build/build.rs
aPere3: Oh yeah, you must be right ! I did it without thinking to be honest ! Good catch !

concrete-core-wasm/build/build.rs
aPere3: Done !

concrete-core-wasm/build/build.rs
aPere3: Done

concrete-core-wasm/build/build.rs
aPere3: At this moment the file is not yet written ! At first I wrote the file and then called rustfmt, but then moved to this solution when I thought I would need to inline stuffs for wasm bindgen.

Do you think it is worth getting back to writing + rustfmt on file ?

concrete-core-wasm/build/prune.rs
aPere3: Hmmm. Not sure what you imagine for that. The two things I can think of to enhance that are:
+ Replace it by a static whitelist.
+ Generate a dynamic whitelist based on feature flags. For now we would just have `default` ... Oh wait, yes and `fft` also probably !! We could have features `backend_fft` and `backend_default`, detect that and create a whitelist based on those ?

concrete-core-wasm/build/build.rs
IceTDrinker: I'm not used to writing directly to stdin, up to you !

Potentially if we call rustfmt like when we format on our dev machines it can format everything in place, really up to you on this one !

concrete-core-wasm/src/lib.rs
aPere3: Because those are not part of the ccr per se. There is no `Parameter` node in the ccr because well, I did not do it as discussed in the other PR. That being said, we will be able to generate those very quickly when the parameters are back into `concrete-core`.

IceTDrinker: thx !

concrete-core-wasm/build/prune.rs
IceTDrinker: may be overkill for now but yeah I think having some backend features could be nice 🙂

concrete-core-wasm/build/generation.rs
IceTDrinker: ok !

agnesLeroy:
@slab-ci cpu_test

IceTDrinker (approved):
Ok for me, should we merge to solve the build issue we are getting with concrete-core-wasm and clippy @aPere3 ?

PR #160, opened 2022-09-05
feat(concrete-fft): add concrete-fft backend

sarah-ek:
### Description
this PR intergrates a new pure rust fft backend. right now the dependency points to the concrete-fft github repo. but this should be changed to a crates.io version once the crate is published.

### Checklist 

### DIFF
--- a/concrete-commons/src/numeric/mod.rs
+++ b/concrete-commons/src/numeric/mod.rs
@@ -21,7 +21,7 @@ mod signed;
 mod unsigned;
 
 /// A trait implemented by any generic numeric type suitable for computations.
-pub trait Numeric: Sized + Copy + PartialEq + PartialOrd {
+pub trait Numeric: Sized + Copy + PartialEq + PartialOrd + 'static {
     /// This size of the type in bits.
     const BITS: usize;
 
--- a/concrete-core-bench/Cargo.toml
+++ b/concrete-core-bench/Cargo.toml
@@ -17,6 +17,11 @@ backend_fftw = [
     "concrete-core/backend_fftw",
     "concrete-core-fixture/backend_fftw",
 ]
+backend_fft = [
+    "concrete-core/backend_fft",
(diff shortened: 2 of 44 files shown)


agnesLeroy:
@sarah-ek let's wait for the concrete-fft repository to go public and for the crate to be pushed to crates.io: otherwise we can't verify the CI is passing.

agnesLeroy:
@sarah-ek your message commit doesn't comply with the convention we check for, you can use this instead:
```
feat(core): ...
```

agnesLeroy (approved):

IceTDrinker:
@slab-ci cpu_test

agnesLeroy:
@sarah-ek do yo uhave an idea what's causing the build error?

agnesLeroy:
@sarah-ek for what you can do is use the compilation flag `__ci_do_not_compile` to remove that feature when `--all-features` is used. This is what's used for the Cuda support in the CI. Then you'll have to add a task for the nightly feature (in concrete-tasks and in the workflow file for slab) -> we don't use the labels to trigger workflows so only updating the slab workflow should be good.

sarah-ek:
this might be an issue because `backend_fft_nightly_avx512` also enables nightly features in `concrete-fft`. i don't think we can have a feature that disables that since cargo expects features to be additive.

would it be possible to add an `__all_stable_features` feature that enables all stable features, and change the CI from using `--all-features` to something like `--features=__al_stablel_features`?

IceTDrinker:
feature handling will change with the aarch64 PR, I think it's fine to have a band aid fix for now and we'll manage it properly with the addition of the aarch64 macs support, wdyt ?

sarah-ek:
for now i commented out the simd features. we can enable them again once we have figured out a solution for the CI

agnesLeroy:
@slab-ci cpu_test

PR #186, opened 2022-09-15
feat(core): support for M1 macs

IceTDrinker:
### Description

PR to make our crates work on apple M1 and generally aarch64

I will deduplicate the code for ffi release, but felt like this was worth pushing

### Checklist 

### DIFF
--- a/.github/workflows/ffi_github_release.yml
+++ b/.github/workflows/ffi_github_release.yml
@@ -8,7 +8,6 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  OUTPUT_BUILD_DIR: "target/release/"
 
 jobs:
   build-linux:
@@ -22,69 +21,18 @@ jobs:
       - uses: actions/checkout@v3
       - name: Build libconcrete_core_ffi
         env:
-          RUSTFLAGS: "-Ctarget-feature=+aes,+sse2"
+          RUSTFLAGS: "-Ctarget-feature=+aes,+sse2,+avx,+avx2"
         run: |
-          cargo build --release \
-          --features backend_default \
-          --features backend_default_generator_x86_64_aesni \
-          --features backend_default_serialization \
-          --features backend_default_parallel \
-          --features backend_fftw \
-          --features backend_fftw_serialization \
-          --features seeder_unix \
-          --features seeder_x86_64_rdseed \
-          -p concrete-core-ffi
-
-          echo "Build artifacts:"
-
-          # Check the build output
-          ls "${OUTPUT_BUILD_DIR}"
-
-          # Use cargo metadata to emit all metadata in json form, format version is recommended
-          # for compatibility. Then use the jq (r: raw string, c: condensed) filter which does:
-          # Iterate over json["packages"] as array, select the package named "concrete-core-ffi" and
-          # output its version
-          CONCRETE_CORE_FFI_VERSION="$(cargo metadata --format-version 1 |
-            jq -rc '.packages[] | select(.name=="concrete-core-ffi") | .version')"
-
-          echo "concrete-core-ffi version: ${CONCRETE_CORE_FFI_VERSION}"
-
-          # Create a temporary directory and a raw artifacts dir in it
           TMP_DIR="$(mktemp -d)"
-          RAW_ARTIFACTS_DIR="${TMP_DIR}/raw_artifacts"
-          mkdir -p "${RAW_ARTIFACTS_DIR}"
 
-          echo "TO_UPLOAD=${TMP_DIR}/*" >> "${GITHUB_ENV}"
-
-          # Copy the build artifacts to the raw_artifacts dir
-          # Hard code the names of the files we want to archive
-          cp "${OUTPUT_BUILD_DIR}/concrete-core-ffi.h" "${RAW_ARTIFACTS_DIR}/"
-          cp "${OUTPUT_BUILD_DIR}/libconcrete_core_ffi.a" "${RAW_ARTIFACTS_DIR}/"
-          cp "${OUTPUT_BUILD_DIR}/libconcrete_core_ffi.so" "${RAW_ARTIFACTS_DIR}/"
-
-          OUTPUT_TARBALL="${TMP_DIR}/concrete-core-ffi_${CONCRETE_CORE_FFI_VERSION}_linux_amd64.tar.gz"
-          OUTPUT_ZIP="${TMP_DIR}/concrete-core-ffi_${CONCRETE_CORE_FFI_VERSION}_linux_amd64.zip"
-
-          # Goto TMP_DIR first then RAW_ARTIFACTS_DIR
-          pushd "${TMP_DIR}"
-          pushd "${RAW_ARTIFACTS_DIR}"
-
-          # Create archives
-          echo "Creating tarball..."
-          tar -cvzf "${OUTPUT_TARBALL}" ./*
-
-          echo "Creating zip archive..."
-          zip -r "${OUTPUT_ZIP}" ./*
-
-          # Return to TMP_DIR
-          popd
-          # Create the sha256 for both archives
-          ls concrete-core-ffi* | sort | xargs shasum -a 256 -b > sha256_linux.txt
-
-          # Remove the copies to make our lives easier when packaging artifacts
-          rm -rf "${RAW_ARTIFACTS_DIR}"
-
-          ls
+          ./script/action_utils/concrete-core-ffi-release.sh \
+          --rust-toolchain "$(cargo xtask get_build_toolchain | tail -n 1)" \
+          --features-string '--features x86_64' \
+          --rust-build-dir "$(pwd)/target/release" \
+          --output-release-dir "${TMP_DIR}" \
+          --sha256-sum-out-file sha256_linux.txt \
+          --github-env-file "${GITHUB_ENV}" \
+          --release-flavor linux_amd64
 
       - name: Upload build artifacts
         uses: actions/upload-artifact@v3
@@ -103,72 +51,59 @@ jobs:
       - uses: actions/checkout@v3
       - name: Build libconcrete_core_ffi
         env:
-          RUSTFLAGS: "-Ctarget-feature=+aes,+sse2"
+          RUSTFLAGS: "-Ctarget-feature=+aes,+sse2,+avx,+avx2"
         run: |
-          cargo build --release \
-          --features backend_default \
-          --features backend_default_generator_x86_64_aesni \
-          --features backend_default_serialization \
-          --features backend_default_parallel \
-          --features backend_fftw \
-          --features backend_fftw_serialization \
-          --features seeder_unix \
-          --features seeder_x86_64_rdseed \
-          -p concrete-core-ffi
-
-          echo "Build artifacts:"
-
-          # Check the build output
-          ls "${OUTPUT_BUILD_DIR}"
-
-          # Use cargo metadata to emit all metadata in json form, format version is recommended
-          # for compatibility. Then use the jq (r: raw string, c: condensed) filter which does:
-          # Iterate over json["packages"] as array, select the package named "concrete-core-ffi" and
-          # output its version
-          CONCRETE_CORE_FFI_VERSION="$(cargo metadata --format-version 1 |
-            jq -rc '.packages[] | select(.name=="concrete-core-ffi") | .version')"
-
-          echo "concrete-core-ffi version: ${CONCRETE_CORE_FFI_VERSION}"
-
-          # Create a temporary directory
           TMP_DIR="$(mktemp -d)"
-          RAW_ARTIFACTS_DIR="${TMP_DIR}/raw_artifacts"
-          mkdir -p "${RAW_ARTIFACTS_DIR}"
 
-          echo "TO_UPLOAD=${TMP_DIR}/*" >> "${GITHUB_ENV}"
-
-          # Copy the build artifacts to the raw_artifacts dir
-          # Hard code the names of the files we want to archive
-          cp "${OUTPUT_BUILD_DIR}/concrete-core-ffi.h" "${RAW_ARTIFACTS_DIR}/"
-          cp "${OUTPUT_BUILD_DIR}/libconcrete_core_ffi.a" "${RAW_ARTIFACTS_DIR}/"
-          cp "${OUTPUT_BUILD_DIR}/libconcrete_core_ffi.dylib" "${RAW_ARTIFACTS_DIR}/"
-
-          OUTPUT_TARBALL="${TMP_DIR}/concrete-core-ffi_${CONCRETE_CORE_FFI_VERSION}_darwin_amd64.tar.gz"
-          OUTPUT_ZIP="${TMP_DIR}/concrete-core-ffi_${CONCRETE_CORE_FFI_VERSION}_darwin_amd64.zip"
-
-          # Goto TMP_DIR first then RAW_ARTIFACTS_DIR
-          pushd "${TMP_DIR}"
-          pushd "${RAW_ARTIFACTS_DIR}"
-
-          # Create archives
-          echo "Creating tarball..."
-          tar -cvzf "${OUTPUT_TARBALL}" ./*
-
-          echo "Creating zip archive..."
-          zip -r "${OUTPUT_ZIP}" ./*
-
-          # Return to TMP_DIR
-          popd
-          # Create the sha256 for both archives
-          ls concrete-core-ffi* | sort | xargs shasum -a 256 -b > sha256_darwin.txt
-
-          # Remove the copies to make our lives easier when packaging artifacts
-          rm -rf "${RAW_ARTIFACTS_DIR}"
-
-          ls
+          ./script/action_utils/concrete-core-ffi-release.sh \
+          --rust-toolchain "$(cargo xtask get_build_toolchain | tail -n 1)" \
+          --features-string '--features x86_64' \
+          --rust-build-dir "$(pwd)/target/release" \
+          --output-release-dir "${TMP_DIR}" \
+          --sha256-sum-out-file sha256_darwin.txt \
+          --github-env-file "${GITHUB_ENV}" \
+          --release-flavor darwin_amd64
 
       - name: Upload build artifacts
         uses: actions/upload-artifact@v3
         with:
           name: darwin_amd64
+          path: ${{ env.TO_UPLOAD }}
+
+  build-mac-os-apple-silicon:
+    name: Build concrete-core-ffi on macOS with M1
+    runs-on: ["self-hosted", "m1mac"]
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - name: Install Rust for tasks
+        run: |
+          make install_tasks_rust_toolchain
+      - name: Build libconcrete_core_ffi
+        env:
+          RUSTFLAGS: "-Ctarget-cpu=native"
+        run: |
+          TMP_DIR="$(mktemp -d)"
+
+          ./script/action_utils/concrete-core-ffi-release.sh \
+          --rust-toolchain "$(cargo xtask get_build_toolchain | tail -n 1)" \
+          --features-string '--features aarch64' \
+          --rust-build-dir "$(pwd)/target/release" \
+          --output-release-dir "${TMP_DIR}" \
+          --sha256-sum-out-file sha256_darwin.txt \
+          --github-env-file "${GITHUB_ENV}" \
+          --release-flavor darwin_aarch64
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: darwin_aarch64
           path: ${{ env.TO_UPLOAD }}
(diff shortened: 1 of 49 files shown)


IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci gpu_test

concrete-tasks/src/utils.rs
```diff
@@ -62,10 +62,57 @@ pub fn get_nightly_toolchain() -> Result<String, Error> {
...
     let toolchain_arg = format!("+{}", toolchain_base);
 
-    if cfg!(target_os = "macos") {
-        // For now we don't support Apple Silicon, always use the x86_64 toolchain for tasks
-        Ok(format!("{}-x86_64-apple-darwin", toolchain_arg))
+    Ok(toolchain_arg)
+}
+
+pub fn get_build_toolchain() -> Result<String, Error> {
+    if cfg!(target_arch = "aarch64") && cfg!(target_os = "macos") {
```
IceTDrinker, line 69: not correct, should for aarch64 in general

concrete-tasks/src/utils.rs
```diff
@@ -62,10 +62,57 @@ pub fn get_nightly_toolchain() -> Result<String, Error> {
...
+            Err(Error::new(
+                ErrorKind::Other,
+                "Unsupported target_arch and target_os combination",
+            ))
+        }
     } else {
-        Ok(toolchain_arg)
+        Err(Error::new(ErrorKind::Other, "Unsupported target_arch"))
     }
 }
```
IceTDrinker, line 91: should be adapted to fail less spectacularly on other platforms like windows as we will now be windows compatible by default

IceTDrinker: thanks to the pure rust fft backend

concrete-tasks/src/build.rs
```diff
@@ -38,86 +40,106 @@ pub mod debug {
...
         } else if cfg!(target_os = "macos") {
-            cmd!(<ENV_TARGET_NATIVE> "cargo test --release --no-run --all-features")
+            &ENV_TARGET_NATIVE
         } else {
             unreachable!()
-        }
-    }
+        };
 
-    pub fn doctests() -> Result<(), Error> {
```
IceTDrinker, line 112: not sure why doctests were removed (probably never used in our code, and the cargo test does run doctests as well)

IceTDrinker:
@slab-ci gpu_test

concrete-core-fixture/src/raw/generation.rs
```diff
@@ -1,24 +1,32 @@
...
+#[cfg(feature = "fixture_generator_x86_64_aesni")]
 use concrete_csprng::generators::AesniRandomGenerator;
-#[cfg(not(target_arch = "x86_64"))]
+#[cfg(feature = "fixture_generator_aarch64_aes")]
+use concrete_csprng::generators::ArmAesRandomGenerator;
+#[cfg(all(
+    not(feature = "fixture_generator_x86_64_aesni"),
+    not(feature = "fixture_generator_aarch64_aes")
+))]
 use concrete_csprng::generators::SoftwareRandomGenerator;
```
IceTDrinker, line 12: generator conf can probably be put in a separate file like was done in core

IceTDrinker:
@slab-ci gpu_test

sarah-ek:
i can't comment much on the CI stuff. but the code changes look good to me

soonum (commented):
Thanks Arthur for the effort you've put to support M1 architecture.
From a CI standpoint everything looks fine to me.

concrete-csprng/Cargo.toml
```diff
@@ -13,7 +13,10 @@ keywords = ["fully", "homomorphic", "encryption", "fhe", "cryptography"]
 
 [dependencies]
 aes-soft = "0.6.4"
-rayon = {version="1.5.0", optional= true}
+rayon = { version = "1.5.0", optional = true }
+
+[target.aarch64-apple-darwin.dependencies]
+libc = "0.2.107"
```
tmontaigu, line 19: Since the PR does not include the apple enclave seeder, this is not needed I think

script/make_utils/install_rust_toolchain.sh
```diff
@@ -39,11 +39,6 @@ do
    shift
 done
 
-if [[ "$(uname)" == "Darwin" ]]; then
-    # We are on mac OS, specify the x86_64 toolchain for now
-    TOOLCHAIN_VERSION="${TOOLCHAIN_VERSION}-x86_64-apple-darwin"
-fi
-
```
tmontaigu, line 46: I think the `usage()` function should be updated since you don't always install the x86_64 toolchain on apple

tmontaigu (commented):
I'm confused, does this require the nightly toolchain to compile of Apple Silicon Mac or the stable one works ?

script/make_utils/install_rust_toolchain.sh
IceTDrinker: you are right

concrete-csprng/Cargo.toml
IceTDrinker: correct

IceTDrinker:
> I'm confused, does this require the nightly toolchain to compile on Apple Silicon Mac or the stable one works ?

To be able to use the accelerated AES generator yes it requires the nightly toolchain for apple M1

IceTDrinker:
there are still a few aarch64 intrisics that are not marked as stable

concrete-core/src/backends/fftw/private/math/fft/polynomial.rs
```diff
@@ -342,7 +342,10 @@ fn regular_uhwap<C1, C2, C3>(
     }
 }
 
-#[allow(unused)]
+#[cfg(all(
+    any(target_arch = "x86", target_arch = "x86_64"),
+    target_feature = "avx2",
+))]
```
agnesLeroy, line 348: We had missed this in @sarah-ek's PR?

.github/workflows/ffi_github_release.yml
```diff
@@ -22,17 +22,10 @@ jobs:
       - uses: actions/checkout@v3
       - name: Build libconcrete_core_ffi
         env:
-          RUSTFLAGS: "-Ctarget-feature=+aes,+sse2"
+          RUSTFLAGS: "-Ctarget-feature=+aes,+sse2,+avx,+avx2"
```
agnesLeroy, line 25: Why do you have to add these here?

agnesLeroy (commented):
@IceTDrinker that's some piece of work! :tada: Thanks a lot @tmontaigu for the initial work on this as well. :smile: Here come a few comments of mine, nothing blocking you'll see. Cheers!

concrete-core-fixture/src/raw/generation.rs
agnesLeroy: If we only use it here I'd say we can leave it as it is for now :slightly_smiling_face:

concrete-tasks/src/utils.rs
IceTDrinker: hmm yeah you are most likely right for windows, though there are some weird arm windows versions.

As for the else part, it's not correct to say we don't support the arch as most code with fft can now be compiled for any supported rust target

concrete-tasks/src/build.rs
IceTDrinker: yes it's me, I just don't remember why I did it, but I think it's because we don't use it ever in CI, and I just tend to write my own cargo commands to test what I want, so yeah if it's fine I'll just remove them !

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
```diff
@@ -0,0 +1,173 @@
...
+};
+use std::arch::is_aarch64_feature_detected;
+use std::mem::transmute;
+
+const RCONS: [u32; 10] = [0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80, 0x1B, 0x36];
+const NUM_WORDS_IN_KEY: usize = 4;
+const NUM_ROUNDS: usize = 10;
+const NUM_ROUND_KEYS: usize = NUM_ROUNDS + 1;
+
+/// An aes block cipher implementation which uses `neon` and `aes` instructions.
```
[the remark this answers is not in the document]
IceTDrinker: what do you mean even better support ?

The implementation was made with intrisics of the arm chip, maybe there is a way to write it differently but as I'm not the author of this code I can't say

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
```diff
@@ -0,0 +1,173 @@
...
+    let mut data: uint8x16_t = u128_to_uint8x16_t(message);
+
+    for &key in keys.iter().take(NUM_ROUNDS - 1) {
+        data = vaesmcq_u8(vaeseq_u8(data, key));
+    }
+
+    data = vaeseq_u8(data, keys[NUM_ROUNDS - 1]);
+    data = veorq_u8(data, keys[NUM_ROUND_KEYS - 1]);
+
+    uint8x16_t_to_u128(data)
```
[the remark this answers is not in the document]
IceTDrinker: I think there are tests checking the correctness in the same file ?

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
```diff
@@ -0,0 +1,173 @@
...
+        if (i % NUM_WORDS_IN_KEY) == 0 {
+            words[i] = words[i - NUM_WORDS_IN_KEY]
+                ^ sub_word(words[i - 1]).rotate_right(8)
+                ^ RCONS[(i / NUM_WORDS_IN_KEY) - 1];
+        } else {
+            words[i] = words[i - NUM_WORDS_IN_KEY] ^ words[i - 1];
+        }
+        // Note: there is also a special thing to do when
+        // i mod SElf::NUM_WORDS_IN_KEY == 4 but it cannot happen on 128 bits keys
+    }
```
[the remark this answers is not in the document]
IceTDrinker: we could check the different generators against each other over a number of bytes, the software one comes from the aes crate so hopefully is correct 😅

concrete-tasks/src/test.rs
```diff
@@ -14,31 +17,61 @@ lazy_static! {
...
+        &format!("cargo {} test --release --no-fail-fast --features {} -p concrete-csprng",
+        get_build_toolchain()?,
+        get_target_arch_feature_for_core()?
+    ))
 }
 
 pub fn npe() -> Result<(), Error> {
-    cmd!(<ENV_TARGET_NATIVE> "cargo test --release --no-fail-fast --all-features -p concrete-npe")
+    cmd!(<ENV_TARGET_NATIVE>
+        &format!("cargo {} test --release --no-fail-fast --all-features -p concrete-npe",
```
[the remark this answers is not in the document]
IceTDrinker: yeah, npe is a bit weird and does not really have anything arch specific, though it has 0 features, so I can remove the --all-features here !

concrete-tasks/src/utils.rs
```diff
@@ -62,10 +62,57 @@ pub fn get_nightly_toolchain() -> Result<String, Error> {
...
+        get_target_arch_feature_for_core()?
+    ))
+}
+
+pub fn get_target_arch_feature_flags_for_all_crates() -> Result<String, Error> {
+    Ok(format!(
+        "--features {} --features {} --features {}",
+        get_target_arch_feature_for_core()?,
+        get_target_arch_feature_for_tests()?,
+        get_target_arch_feature_for_benches()?
```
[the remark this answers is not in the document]
IceTDrinker: not exactly, but maybe it wasn't smart on my end to have an "optimized_benches_for_x86_64", I take suggestion on how to name all these features

concrete-core/src/backends/fftw/private/math/fft/polynomial.rs
IceTDrinker: no, don't worry, it's because on aarch64 it triggers some errors on clippy checks as it cannot use x86 intrisics, so I feature/target gate it 🙂

.github/workflows/ffi_github_release.yml
IceTDrinker: I don't have to, but basically currently we don't allow rust to emit avx/avx2 instructions which are vector intrisics for the rust code as the x86 version needs to run m1 macs in emulation mode (which does not have avx/avx2). Now that we have a proper M1 support we can decorrelate the features we enable for the FFI from the M1 requirements, we discussed a while back that avx/avx2 was fair game to use, so I took the opportunity to enable them 🙂 

We can potentially enable more but we need to see what CPUs support, I think avx/avx2 is nearly ubiquitous

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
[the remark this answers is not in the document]
agnesLeroy: Oh ok, it's just I thought there existed other instructions, but I have no idea whether they would be useful for us. Maybe @tmontaigu you know better?

concrete-tasks/src/utils.rs
agnesLeroy: So we could add a warning stating that the returned toolchain will not necessarily work out for the platform & set of features they use.

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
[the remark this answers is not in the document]
agnesLeroy: Yes, but I guess having the check with the NIST tool will be good as well.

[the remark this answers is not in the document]
IceTDrinker: what I mean is there is an expected good answer for the AES block cipher as it's deterministic

.github/workflows/ffi_github_release.yml
agnesLeroy: Oh alright, nice! Hum, then it'd be good indeed to investigate what features we can enable here that'll speed things up for users.

concrete-core-bench/Cargo.toml
```diff
@@ -12,25 +12,53 @@ paste = "1.0"
...
-[target.'cfg(target_arch = "x86_64")'.dependencies]
-# This does nothing if backend_default is not requested by the user
-concrete-core = { path = "../concrete-core", features = [
+optimized_benches_for_x86_64 = [
+    "concrete-core-fixture/fixture_generator_x86_64_aesni",
+]
+optimized_benches_for_x86_64-cuda = ["optimized_benches_for_x86_64"]
+optimized_benches_for_aarch64 = [
+    "concrete-core-fixture/fixture_generator_aarch64_aes",
+]
```
agnesLeroy, line 48: It seems to me we could keep only the x86_64, x86_64-cuda and aarch64 features: if we're on an x86_64 platform we want to use the x86_64 features in the fixtures, and same on an aarch64 platform. What do you think?

IceTDrinker: fine by me, less features, not sure what the rationale behind those were when I first wrote the PR a few months ago

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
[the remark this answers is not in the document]
aPere3: Actually, this is tested against the test vector of Fips 197. So hopefully this should be correct.

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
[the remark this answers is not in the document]
agnesLeroy: Sure :slightly_smiling_face:

concrete-cuda/Cargo.toml
```diff
@@ -16,7 +16,3 @@ cmake = { version = "0.1" }
 
 [dependencies]
 thiserror = "1.0"
-
-[features]
-# A feature to disable build on github runners for the CI
-_ci_do_not_compile = []
```
aPere3, line 22: I don't really remember why this flag was needed. What makes it possible to remove it with this PR ?

concrete-core/Cargo.toml
```diff
@@ -82,11 +85,35 @@ __commons_serialization = [
...
+    "backend_default",
+    "backend_default_parallel",
+    "backend_default_generator_aarch64_aes",
+    "backend_default_serialization",
+    "backend_fftw",
+    "backend_fftw_serialization",
+    "backend_fft",
+    "backend_fft_serialization",
+    "seeder_unix",
+]
```
aPere3, line 115: If I understand correctly, those are shortcuts features, that activates all public features for some profiles right ? 

Maybe it would be nice to document that on top of the features ?

concrete-core-fixture/Cargo.toml
```diff
@@ -5,28 +5,35 @@ edition = "2021"
...
+]
 backend_default_parallel = ["concrete-core/backend_default_parallel"]
 backend_cuda = ["concrete-core/backend_cuda", "concrete-cuda"]
+backend_default_generator_aarch64_aes = [
+    "concrete-core/backend_default_generator_aarch64_aes",
+]
 
-_ci_do_not_compile = []
+# Feature to have efficient AES implementation for generation in fixtures
+fixture_generator_x86_64_aesni = ["concrete-csprng/generator_x86_64_aesni"]
```
aPere3, line 37: That's great ! :tada:

aPere3 (commented):
Hello @IceTDrinker , @tmontaigu ,

Thanks for this fantastic work, this will be a game changer for many users of the library.

I don't have much to say honestly. This all looks pretty good for me ! 

Best,

Alex

concrete-cuda/Cargo.toml
IceTDrinker: before we used --all-features, now we can use the x86_64-like profiles

concrete-core/Cargo.toml
IceTDrinker: will do

concrete-tasks/src/utils.rs
IceTDrinker: So here in fact it's not the toolchain thing, it is the x86_64 like features I added, so it is indeed a target_arch we don't have a configuration for, I'll add a more explicit message

concrete-tasks/src/utils.rs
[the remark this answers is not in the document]
IceTDrinker: so actually after updating the features in test and bench those additional features should not be required anymore

.github/workflows/ffi_github_release.yml
IceTDrinker: https://github.com/zama-ai/concrete-core-internal/issues/423

concrete-core-bench/Cargo.toml
IceTDrinker: ok so I think we can keep the intermediate benches/test features, I now enable them in the x86_64 and other profiles

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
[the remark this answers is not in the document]
tmontaigu: What kind of other instructions are you thinking about ?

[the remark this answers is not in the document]
agnesLeroy: Well I don't know much about M1 chips, but it seems to me the Neon set of instructions provide aes instructions, and the sve2 one as well, no? https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiessimdisa=[sve2]&q=aes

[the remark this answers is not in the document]
IceTDrinker: sve2 could be better than neon it seems, not sure about their general availability on arm chips and Apple M1 

From https://developer.arm.com/documentation/102340/0001/Introducing-SVE2 :

> SVE and SVE2 are not an extension of the Neon instruction set. Instead, SVE and SVE2 are redesigned for better data parallelism than Neon provides. However, the hardware logic of SVE and SVE2 overlays the Neon hardware implementation. When a microarchitecture supports SVE or SVE2, it also supports Neon. To use SVE and SVE2, software that runs on that microarchitecture must first use Neon.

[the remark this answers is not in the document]
tmontaigu: It is not available on M1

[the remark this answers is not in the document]
IceTDrinker: closing this then

concrete-csprng/src/generators/implem/aarch64/block_cipher.rs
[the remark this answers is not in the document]
IceTDrinker: there is a check with a test vector as indicated by Alex, closing

IceTDrinker:
@slab-ci gpu_test

IceTDrinker:
still need to fix the ffi release script

IceTDrinker:
FFI builds work https://github.com/zama-ai/concrete-core/actions/runs/3083749815

IceTDrinker:
Should we disable m1 tests on PR and leave it as something manual ? (through labels for now, as it would require a lot of work to make slab use the m1 mac from the office, plus github will have M1 instances at some point)

IceTDrinker:
For the m1 tests on every PR we can think about that later, once we have discussed with other teams to put the CI strategy in place, at the moment we will be the only team with an M1 build and end of quarter means we won't see M1 support up in the stack before next quarter at the earliest, so this should be fine for now

agnesLeroy (approved):

PR #190, opened 2022-09-19
feat(core): add private implementation for the NTT backend

blowfish880:
### Description
This adds the private part of the NTT backend. The NTT algorithm, including modular arithmetic and parameter choices, is implemented rather naively and to be improved. 

This [report](https://github.com/zama-ai/concrete-core/files/9801924/main.pdf) provides a description of the structure of the backend.

### Checklist 


agnesLeroy:
@slab-ci cpu_test

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
```diff
@@ -0,0 +1,29 @@
+use crate::backends::ntt::private::math::transform::Ntt;
+use crate::commons::crypto::glwe::GlweCiphertext;
+use crate::commons::math::torus::UnsignedTorus;
+use crate::commons::numeric::UnsignedInteger;
+use crate::prelude::{GlweSize, PolynomialSize};
+
+pub struct FourierBuffers<Scalar, Nntt: UnsignedInteger> {
+    pub ntt: Ntt<Nntt>,
+    // Those buffers are also used to store the lut and the rounded input during the bootstrap.
+    pub lut_buffer: GlweCiphertext<Vec<Scalar>>,
```
agnesLeroy, line 10: The LUT is given as input in a GLWE ciphertext for the bootstrap, why do you need an additional buffer?

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub rounded_buffer: GlweCiphertext<Vec<Scalar>>,
+}
+
+impl<N, Nntt> FourierBuffers<N, Nntt>
+where
+    N: UnsignedTorus,
+    Nntt: UnsignedInteger,
```
agnesLeroy, line 17: Why do you need this Nntt type? It seems there's a typo in the name?

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/mod.rs
```diff
@@ -0,0 +1,718 @@
...
+    /// let bsk: NttBootstrapKey<Vec<ModQ<u64>>> = NttBootstrapKey::allocate(
+    ///     GlweSize(7),
+    ///     PolynomialSize(256),
+    ///     DecompositionLevelCount(3),
+    ///     DecompositionBaseLog(5),
+    ///     LweDimension(4),
+    /// );
+    /// assert_eq!(bsk.key_size(), LweDimension(4));
+    /// ```
+    pub fn key_size(&self) -> LweDimension
```
agnesLeroy, line 343: `key_size` is a legacy wording we're trying to remove from the code: instead you can use input_lwe_dimension I think.

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/mod.rs
```diff
@@ -0,0 +1,718 @@
...
+                NttGgswCiphertext::from_container(
+                    tensor.into_container(),
+                    rlwe_size,
+                    poly_size,
+                    base_log,
+                )
+            })
+    }
+
+    fn blind_rotate<C2, Nntt, Nin>(
```
agnesLeroy, line 457: @IceTDrinker @aPere3 I think we could start exposing those functions in a separate file instead of attaching them to the structure, wdyt?

agnesLeroy (commented):
Hello @blowfish880!Thanks a lot for this contribution, this is great! @sarah-ek can I let you take a look at the PR as well? There'll be a refactor to do in the same fashion as in the rest of core to remove asRefTensor/asMutTensor. Also, the functions are currently attached to the structures but it'd be easier to have them as separate functions (bootstrap, cmux, blind rotate, external product). This refactor could be done in a second step as well.

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
blowfish880: This is following the FFTW backend, where the input accumulator is copied into this buffer. I'm assuming this is done so that the input accumulator does not need to be modified.

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
blowfish880: There are two integer like types involved in this backend: the usual type for the ciphertext mod 2^32, resp. 2^64, and the type parameter for values mod Q, which is what is used internally in the NTT. These two need to be different, since the latter needs to be twice the size of the former in order to have Q correspond roughly to 2^32, resp. 2^64, and still support easy modular multiplication. So, for 32bit ciphertexts, Nntt will be u64 and for 64bit ciphertexts it will be u128. There are probably better names for it and potentially ways to do modular multiplication without using data types with twice the size, but this was the most straight forward way to go.

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
```diff
@@ -0,0 +1,29 @@
+use crate::backends::ntt::private::math::transform::Ntt;
+use crate::commons::crypto::glwe::GlweCiphertext;
+use crate::commons::math::torus::UnsignedTorus;
+use crate::commons::numeric::UnsignedInteger;
+use crate::prelude::{GlweSize, PolynomialSize};
+
+pub struct FourierBuffers<Scalar, Nntt: UnsignedInteger> {
```
IceTDrinker, line 7: is this really a FourierBuffer ?

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub ntt: Ntt<Nntt>,
+    // Those buffers are also used to store the lut and the rounded input during the bootstrap.
+    pub lut_buffer: GlweCiphertext<Vec<Scalar>>,
+    pub rounded_buffer: GlweCiphertext<Vec<Scalar>>,
+}
+
+impl<N, Nntt> FourierBuffers<N, Nntt>
```
IceTDrinker, line 14: Proposal for naming:

Scalar (which used extensively in the commons private API)
NttScalar (to indicate it's the scalar type used to perform NTT computations), you can even add a comment to clear that up

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/mod.rs
```diff
@@ -0,0 +1,718 @@
...
+use crate::commons::math::torus::UnsignedTorus;
+
+use crate::commons::numeric::{CastFrom, CastInto, UnsignedInteger};
+use crate::prelude::{
+    DecompositionBaseLog, DecompositionLevelCount, GlweSize, LutCountLog, LweDimension,
+    ModulusSwitchOffset, MonomialDegree, PolynomialSize,
+};
+pub use buffers::FourierBuffers;
+
+/// A bootstrapping key in the fourier domain.
```
IceTDrinker, line 24: I'm guessing this was copy pasted ?

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/mod.rs
```diff
@@ -0,0 +1,718 @@
...
+// #[cfg_attr(feature = "backend_fftw_serialization", derive(Serialize, Deserialize))]
```
IceTDrinker, line 25: this line can either be removed (if we don't add serialization in this PR) or modified to match the serialization feature for your backend if you already had one

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/mod.rs
```diff
@@ -0,0 +1,718 @@
...
+    /// let roots_inv: Vec<ModQ<u64>> = INVROOTS_32_256
+    ///     .to_vec()
+    ///     .iter()
+    ///     .map(|a| <ModQ<u64>>::new(*a as u64, q))
+    ///     .collect();
+    /// let n_inv = ModQ::new(NINV_32_256, q);
+    /// let mut ntt = Ntt::new(poly_size, log_size, roots, roots_inv, n_inv);
+    /// ntt_bsk.fill_with_forward_fourier(&bsk, &mut ntt);
+    /// ```
+    pub fn fill_with_forward_fourier<InputCont, Nin, Nntt>(
```
IceTDrinker, line 192: naming issue I assume ?

IceTDrinker (commented):
I think some naming pass needs to be done on this PR 🙂 fourier -> ntt mostly it seems

It will reduce the noise we generate with the review comments

IceTDrinker:
I'll wait for some commits to correct naming to avoid generating dozens of uninteresting review comments, re-request when you think you caught them all @blowfish880 !

blowfish880:
@IceTDrinker thanks for your comments. Generally speaking, the naming conventions around FFT/DFT/NTT are somewhat fuzzy in the literature. One can think of the NTT as a generalization of the complex FFT to finite fields, or as the FFT and the NTT as different instantiations of the same algorithm, the DFT. In any case, I consider the NTT to also be a Fourier transform, so terms like "Fourier domain" and "forward Fourier transform" etc still make sense to me in the context of the NTT. That being said, I might not have been entirely consistent with the naming, but not sure if we really want to replace all occurrences of "Fourier" with "NTT". I was attempting to replace everything referred to with "FFT" by "NTT", but keep the "Fourier" terminology. Does that make sense?

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
blowfish880: I like that. Will adopt it. 

As for a clarifying comment, there is one in the implementation of the entities where the types are instantiated, but that's not part of the PR. These two types appear in quite a few places within the private part of the transform. Should there be a comment in each such place?

IceTDrinker: maybe not all over the place 🙂

IceTDrinker: I let you be the judge for the comments we don't have them for scalars e.g. in the commons private IIRC

blowfish880:
Do others have opinions about this? Should we get rid of the term `Fourier` throughout the backend? For example, `Fourier domain` could become `NTT domain`, `forward_fourier` could be `forward_transform` or `forward_ntt`, etc.

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
blowfish880: Maybe `BootstrapBuffers` is more appropriate?

IceTDrinker: I recognize the terminology from fftw, so I understand where it comes from, are those buffers required every time we do an NTT ?

blowfish880: The buffers contain an `Ntt` object (which contains a buffer itself), which is used to perform the NTT. The other two buffers are closer related to the bootstrap, not so much to the NTT itself. You're right, the naming comes from the FFTW backend. There it looks like the naming is a historical artifact: my guess is that at first they only contained the FFT buffers, but then they were probably augmented with the other buffers for efficiency reasons.

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/mod.rs
blowfish880: I thought I had caught them all. Thanks! (I had commented these out at first since I wasn't sure if we wanted to add serialization in this PR or add it later. Then @agnesLeroy and I decided that it's probably easier to add it later.)

IceTDrinker: Once the naming is stabilized for ntt/fourier I'll do another pass this is just the one I caught while starting the first light review

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/buffers.rs
agnesLeroy: Ok thanks :slightly_smiling_face:

concrete-core/src/backends/ntt/private/crypto/bootstrap/ntt/mod.rs
agnesLeroy: you'd rather have fill_with_forward_ntt?

agnesLeroy:
> that's what I would have expected and good for me +1

I agree with @IceTDrinker

agnesLeroy:
@blowfish880 have you been able to take the comments into account? Is there something blocking you for this one?

blowfish880:
Did I catch everything?

IceTDrinker:
Sorry I don't know the status on this PR and what we want to merge for 1.0.0 :|

A tip if you need a review you can re-request it clicking the circling arrows around the top right of your screen:

image

The ones that appear near our github usernames

IceTDrinker:
@slab-ci cpu_test

IceTDrinker (approved):
Last commit looks good in terms of renaming, I launched the tests on this version of the PR but it's going to need a rebase on main a squash of the additional commits

For testing this would need public API implementation to plug into the fixtures, do you have that somewhere ?

blowfish880:
Yes, entities, engines and fixtures are forthcoming in future PRs.

Also, I added a report to the description that gives an overview of the backend.

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

agnesLeroy:
@slab-ci cpu_test

PR #199, opened 2022-09-21
Chore/fix cuda test launch

IceTDrinker:
### Checklist 

### DIFF
--- a/.github/workflows/m1_tests.yml
+++ b/.github/workflows/m1_tests.yml
@@ -18,30 +18,32 @@ env:
 
 jobs:
   run-tests-m1:
-    if: "contains(github.event.label.name, 'm1_test')"
+    # Execute if we are not in a PR, and if we are in a PR, only trigger if the label for m1_test is
+    # set
+    if: "github.event_name != 'pull_request' || contains(github.event.label.name, 'm1_test')"
     concurrency:
       group: ${{ github.ref }}_m1mac
       cancel-in-progress: true
     name: Test code in M1
     runs-on: ["self-hosted", "m1mac"]
     steps:
-    - uses: actions/checkout@v3
-    - name: Install Rust
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        override: true
-    - name: Install Rust for tasks
-      run: |
-        make install_tasks_rust_toolchain
-    - name: Run checks
-      run: |
-        cargo xtask check_fmt
-        cargo xtask check_clippy
-        cargo xtask check_doc
-    - name: Run concrete tests
-      run: |
-        cargo xtask test
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - name: Install Rust for tasks
+        run: |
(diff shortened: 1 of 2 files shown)


IceTDrinker:
@slab-ci gpu_test

concrete-tasks/src/test.rs
```diff
@@ -59,7 +59,7 @@ pub fn ffi() -> Result<(), Error> {
 pub fn cuda_test() -> Result<(), Error> {
     cmd!(<ENV_TARGET_NATIVE>
         &format!("cargo test --release --no-fail-fast --features=backend_cuda --features {} \
-        -p concrete-core-test -- --test-threads 1",
+        -p concrete-core-test -- cuda:: --test-threads 1",
```
soonum, line 62: I'm wondering what does a `foo::` option in the tests. It's pure curiosity.

.github/workflows/m1_tests.yml
```diff
@@ -18,30 +18,32 @@ env:
 
 jobs:
   run-tests-m1:
-    if: "contains(github.event.label.name, 'm1_test')"
+    # Execute if we are not in a PR, and if we are in a PR, only trigger if the label for m1_test is
+    # set
+    if: "github.event_name != 'pull_request' || contains(github.event.label.name, 'm1_test')"
```
soonum, line 23: You're reducing the number of cases when the build is triggered, that's right ? you don't want to launch a build on a pending request, I get it. But isn't it mean that any event that trigger this action, other than a pull request, will trigger the build ?

soonum (commented):
Hey @IceTDrinker, looks good to me, I just need a clarification on job triggering.

concrete-tasks/src/test.rs
IceTDrinker: filters the tests tu only run cuda tests

.github/workflows/m1_tests.yml
IceTDrinker: no as the events we authorize are limited, the problem with the current way it is written is that it can only trigger if the event has a label, so the nightly build did not start 🙃

concrete-tasks/src/test.rs
agnesLeroy: I think it's also necessary to restrict in the concrete-core tests, to backends::cuda, otherwise all CPU doc tests will be launched as well.

.github/workflows/m1_tests.yml
soonum: Makes sense then.

soonum (approved):

concrete-tasks/src/test.rs
IceTDrinker: you already did it well for these tests it seems Agnès 🙂

IceTDrinker:
gpu test https://github.com/zama-ai/concrete-core/actions/runs/3096005051/jobs/5011048114

concrete-tasks/src/test.rs
agnesLeroy: Sometimes I do things right :sun_with_face:

IceTDrinker:
M1 test https://github.com/zama-ai/concrete-core/actions/runs/3096005865

PR #207, opened 2022-09-22
feat(core): add AppleSecureEnclaveSeeder in prelude and in FFI

IceTDrinker:
### Description

Make the AppleSecureEnclaveSeeder available in core's prelude and add entry points in the FFI

### Checklist 


IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

concrete-core/src/prelude.rs
```diff
@@ -36,6 +36,8 @@ pub use super::specification::*;
 // --------------------------------------------------------------------------------- CSPRNG REEXPORT
 // Re-export the different seeders of the `concrete-csprng` crate, which are needed to construct
 // default engines.
+#[cfg(target_os = "macos")]
```
agnesLeroy, line 39: Here I thought we would use a feature seeder_aarch64_enclave to activate or not this seeder, instead of relying on the platform, just like for the rdseed seeder. Why did you put the gate on the target OS directly?

agnesLeroy (commented):
Hello! Thanks a lot for this! Here comes my review. We'll need to add an example of use of the Enclave seeder in the user doc in the quick start page.

concrete-core/src/prelude.rs
IceTDrinker: It felt awkward to feature gate rather than use the target OS :| also it is supposed to be available on all mac OS people will use.

I can change it but for this one it feels a bit weird to have a feature flag

agnesLeroy: Alright, for rdseed we had made the feature gate because the feature is not available on all x86_64 platforms, but if all Macs have it then we can indeed just detect the platform.

agnesLeroy (approved):

PR #209, opened 2022-09-24
Cuda/wopbs/cmuxtree

pdroalves:
### Description
This commit replaces device's memory allocation for concrete-cuda's CMUX tree by the equivalent asynchronous methods. This way we can increase efficiency.

### Checklist 


pdroalves:
@slab-ci gpu_test

concrete-cuda/cuda/src/bootstrap_wop.cuh
```diff
@@ -383,20 +383,19 @@ void host_cmux_tree(
...
-    checkCudaErrors(cudaDeviceSynchronize());
+            cudaMemcpyDeviceToDevice, *stream));
 
     // Free memory
-   checkCudaErrors(cudaFree(d_ggsw_fft_in));
-   checkCudaErrors(cudaFree(d_buffer1));
-   checkCudaErrors(cudaFree(d_buffer2));
+   checkCudaErrors(cudaFreeAsync(d_ggsw_fft_in, *stream));
+   checkCudaErrors(cudaFreeAsync(d_buffer1, *stream));
+   checkCudaErrors(cudaFreeAsync(d_buffer2, *stream));
```
agnesLeroy, line 394: @pdroalves I think in the  bootstrap & keyswitch we use cudaFree, and not the async equivalent. Do you know if it makes a lot of difference? Should we change for them too?

agnesLeroy (commented):
Hello @pdroalves! Thanks a lot for this PR, it should speed things up indeed :slightly_smiling_face: I'm just wondering about the async free: we don't use it in the rest of the implementation, and I'm wondering whether we can have it async or if that could cause issues, and what is the impact on performance.

concrete-cuda/cuda/src/bootstrap_wop.cuh
pdroalves: As long as we have a call to cudaStreamSynchronize in the end, it should be enough to assert compatibility to everything else we have in the crate.

I'm running some benchmarks to confirm there are performance improvements by using this strategy. If so, maybe we should do the same there.

pdroalves:
I ran some benchmarks on my laptop (running a RTX 3070) and on our V100 instance and could measure very considerable  speed-ups. On my machine I found numbers between 10% and 46% for the larger instances. On the V100 I measured up to 3.75x !!!

The problem is: looks like this stream-ordered memory allocator is only supported on CUDA >= 11.2. Would it be a problem to set that as our minimum toolkit version?

agnesLeroy:
@pdroalves what will happen if the user has an older version installed? Will it crash, or will the performance just be worse?

pdroalves:
I just tried to compile with cuda 11.1 and got several `undefined reference to cudaMallocAsync`.
[We can use macros](https://stackoverflow.com/questions/37970880/how-to-get-cuda-toolkit-version-at-compile-time-without-nvcc) to select between cudaMalloc and cudaMallocAsync if we want to support this strategy and still offer support to older versions.

```
#if (CUDART_VERSION < 11020)
cudaMalloc((void **)&d_buffer1, num_lut * glwe_size * sizeof(Torus));
#else 
cudaMallocAsync((void **)&d_buffer1, num_lut * glwe_size * sizeof(Torus), *stream);
#endif
```

We have to synchronize the stream at the end anyway, and everything else related to that context will be running in the same stream, so this shouldn't affect correctness. If supported, the asynchronous version will be used and everything will be faster. Otherwise, it will work but slower. 

By using these macros code become less readable, so we can think in a way to encapsulate it.

pdroalves:
@slab-ci gpu_test

concrete-cuda/cuda/src/bootstrap_wop.cuh
```diff
@@ -383,20 +398,30 @@ void host_cmux_tree(
...
             glwe_out, output,
             (glwe_dimension+1) * polynomial_size * sizeof(Torus),
-            cudaMemcpyDeviceToDevice));
+            cudaMemcpyDeviceToDevice, *stream));
 
-    checkCudaErrors(cudaDeviceSynchronize());
+    // We only need synchronization to assert that data is in glwe_out before
+    // returning. Memory release can be added to the stream and processed
+    // later.
+    checkCudaErrors(cudaStreamSynchronize(*stream));
```
agnesLeroy, line 409: Here you exchange the memcopy and the stream sync? And there's no more device sync? Is it normal?

agnesLeroy (commented):
Hello! @pdroalves, looks good to me, there's just a change I am wondering about. :slightly_smiling_face:

concrete-cuda/cuda/src/bootstrap_wop.cuh
pdroalves: Yep, we only need synchronization within the stream, which is cheaper than a full-device synchronization.

agnesLeroy:
@pdroalves https://github.com/zama-ai/concrete-core/pull/214 and https://github.com/zama-ai/concrete-core/pull/215 have to be merged first, then we can re-launch the tests.

pdroalves:
@slab-ci gpu_test

pdroalves:
@slab-ci cpu_test

agnesLeroy (approved):

PR #211, opened 2022-09-27
Add extract bits feature in concrete-cuda

bbarbakadze:
### Description
  implementation of extract bits functionality in concrete-cuda

  update of decomposition algorithm for cuda keyswitch 

### Checklist 

### DIFF
--- a/concrete-cuda/cuda/src/bootstrap_wop.cuh
+++ b/concrete-cuda/cuda/src/bootstrap_wop.cuh
@@ -6,9 +6,7 @@
 #include "../include/helper_cuda.h"
 #include "bootstrap.h"
 #include "complex/operations.cuh"
-#include "crypto/gadget.cuh"
 #include "crypto/torus.cuh"
-#include "crypto/ggsw.cuh"
 #include "fft/bnsmfft.cuh"
 #include "fft/smfft.cuh"
 #include "fft/twiddles.cuh"
@@ -18,6 +16,9 @@
 #include "polynomial/polynomial_math.cuh"
 #include "utils/memory.cuh"
 #include "utils/timer.cuh"
+#include "keyswitch.cuh"
+#include "bootstrap_low_latency.cuh"
+#include "crypto/ggsw.cuh"
 
 template <typename T, class params>
 __device__ void fft(double2 *output, T *input){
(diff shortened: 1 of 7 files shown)


concrete-cuda/cuda/src/bootstrap_wop.cuh
```diff
@@ -399,4 +400,189 @@ void host_cmux_tree(
...
+      dst_copy[params::degree] = src[params::degree];
+      dst_shift[params::degree] = src[params::degree] * value;
+    }
+}
+
+
+// only works for small lwe in ks+bs case
+// function copies lwe when length is not a power of two
+template <typename Torus>
+__global__ void copy_small_lwe(Torus *dst, Torus *src, uint32_t small_lwe_size,
```
agnesLeroy, line 431: what do you mean by "small" here? Is it an LWE encrypting a binary value?

concrete-cuda/cuda/src/bootstrap_wop.cuh
```diff
@@ -399,4 +400,189 @@ void host_cmux_tree(
...
+      cur_dst[tid] = src[tid];
+
+
+}
+
+
+// only used in extract bits for one ciphertext
+// should be called with one block and one thread
+// NOTE: check if putting this functionality in copy_small_lwe or
+// fill_pbs_lut vector is faster
```
agnesLeroy, line 456: If possible it'd seem more efficient to put this together with another kernel indeed, even though it probably won't make a significant difference on the performance.

concrete-cuda/cuda/src/bootstrap_wop.cuh
```diff
@@ -399,4 +400,189 @@ void host_cmux_tree(
...
+         1ll << (ciphertext_n_bits - delta_log - 1));
+
+    for (int bit_idx = 0; bit_idx < number_of_bits; bit_idx++) {
+        cuda_keyswitch_lwe_ciphertext_vector(v_stream, lwe_out_ks_buffer,
+                                             lwe_in_shifted_buffer, ksk,
+                                             lwe_dimension_before,
+                                             lwe_dimension_after, base_log_ksk,
+                                             l_gadget_ksk, 1);
+
+        copy_small_lwe<<<1, 256, 0, *stream>>>(list_lwe_out,
```
agnesLeroy, line 553: Here 256 is hard set: would it be interesting to have it vary depending on the LWE dimension?

concrete-cuda/cuda/src/bootstrap_wop.cuh
```diff
@@ -399,4 +400,189 @@ void host_cmux_tree(
...
+    uint32_t base_log_bsk,
+    uint32_t l_gadget_bsk,
+    uint32_t base_log_ksk,
+    uint32_t l_gadget_ksk)
+{
+    auto stream = static_cast<cudaStream_t *>(v_stream);
+    uint32_t ciphertext_n_bits = sizeof(Torus) * 8;
+
+    int blocks = 1;
+    int threads = params::degree / params::opt;
```
agnesLeroy, line 540: If the degree is 512 the number of threads will be 512/4 = 128 which seems rather small no?

concrete-cuda/cuda/src/crypto/gadget.cuh
```diff
@@ -4,6 +4,7 @@
 #include "polynomial/polynomial.cuh"
 #include <cstdint>
 
+#pragma once
```
agnesLeroy, line 7: Does this make a difference on performance?

concrete-cuda/cuda/src/keyswitch.cuh
```diff
@@ -17,6 +17,18 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
...
+__device__ Torus decompose_one(Torus &state, Torus  mod_b_mask,
+                               int base_log) {
+    Torus res = state & mod_b_mask;
+    state >>= base_log;
+    Torus carry = ((res - 1ll) | state) & res;
+    carry >>= base_log - 1;
+    state += carry;
+    res -= carry << base_log;
+    return res;
+}
```
agnesLeroy, line 30: This is the algorithm that's in concrete-core right now right?

concrete-cuda/cuda/src/keyswitch.cuh
```diff
@@ -76,9 +88,15 @@ __global__ void keyswitch(Torus *lwe_out, Torus *lwe_in,
     Torus a_i = round_to_closest_multiple(block_lwe_in[i], base_log,
                                           l_gadget);
 
+    Torus state = a_i >> (sizeof(Torus) * 8 - base_log * l_gadget);
+    Torus mod_b_mask = (1ll << base_log) - 1ll;
```
agnesLeroy, line 92: Why is a 1ll needed here?

agnesLeroy (commented):
Hello @bbarbakadze! Thanks a lot for this PR! Here come some questions about it :slightly_smiling_face:

concrete-cuda/cuda/src/bootstrap_wop.cuh
bbarbakadze: I believe keyswitch has two modes small to big and big to small, so copying big polynomial is a bit different on cuda since it has the length that is the power of two, and, when we have small lwe, the length might not be the power of two so we need to use a different way to copy, I could make a more generic function just to copy any kind of lwe. Still, like this with small and big lwe it is more optimal.

concrete-cuda/cuda/src/bootstrap_wop.cuh
bbarbakadze: I will measure this during benchmarks. I don't think there will be much difference.

concrete-cuda/cuda/src/keyswitch.cuh
bbarbakadze: Yes it is exactly the same algorithm that is used in concrete-core, and after changing it key switch results match 100% all the time, there were a lot of mismatches before, the key switch itself could pass its own tests, but bit extract was not able to pass the test.

concrete-cuda/cuda/src/keyswitch.cuh
bbarbakadze: to use 64 bit integer

concrete-cuda/cuda/src/bootstrap_wop.cuh
agnesLeroy: Oh alright, so small and big refer to the LWE after & before KS. Could you add a comment above the function just to state that, I think that'd be enough. :slightly_smiling_face: thanks!

concrete-cuda/cuda/src/bootstrap_wop.cuh
agnesLeroy: :+1:

PR #262, opened 2022-11-03
feat(cuda): add support to a larger amount of ciphertexts on the low latency PBS

pdroalves:
### Description
If a non-supported amount of ciphertexts is given to the low latency PBS, splits it in smaller chunks as large as it can handle. 

### Checklist 


pdroalves:
@slab-ci gpu_test

pdroalves:
@slab-ci gpu_test

pdroalves:
@slab-ci gpu_test

pdroalves:
@slab-ci cpu_test

concrete-cuda/cuda/src/bootstrap_low_latency.cuh
```diff
@@ -291,9 +290,23 @@ host_bootstrap_low_latency(void *v_stream, Torus *lwe_array_out,
...
 
-  checkCudaErrors(cudaLaunchCooperativeKernel(
-      (void *)device_bootstrap_low_latency<Torus, params>, grid, thds,
-      (void **)kernel_args, bytes_needed, *stream));
+  int remaining_samples = num_samples;
+  int sample_offset = 0;
+  while(remaining_samples > 0){
+      int num_samples_per_kernel = (
+              2 * remaining_samples * level_count <= cuda_get_number_of_sm(gpu_index)?
+              remaining_samples : cuda_get_number_of_sm(gpu_index) / (2 * level_count));
```
agnesLeroy, line 298: Here you check with `2 * remaining_samples * level_count` but actually since there are several blocks launched per SM shouldn't we take it into account here? So if:
- N = 512 -> 4 blocks per SM are launched, we can check that 2 * remaining_samples * level_count / 4 <= number of SM
- N = 1024, 2048 -> 2 blocks per SM are launched, we can check that  remaining_samples * level_count <= number of SM
What do you think about it?

concrete-cuda/cuda/src/bootstrap_low_latency.cuh
```diff
@@ -258,20 +259,18 @@ host_bootstrap_low_latency(void *v_stream, Torus *lwe_array_out,
                            uint32_t num_samples, uint32_t num_lut_vectors) {
 
   auto stream = static_cast<cudaStream_t *>(v_stream);
+  uint32_t gpu_index = 0; // This must be refactored to support multi-gpu
 
   int buffer_size_per_gpu =
-      level_count * num_samples * polynomial_size / 2 * sizeof(double2);
+      level_count * min(num_samples, cuda_get_number_of_sm(gpu_index)) * polynomial_size / 2 * sizeof(double2);
```
agnesLeroy, line 265: This buffer size should be computed once the number of samples per kernel is known, no? I don't understand why you changed it to `min(num_samples, num_sm)` instead of keeping the same formula but using the number of samples per kernel?

agnesLeroy (commented):
Hi @pdroalves! Thank you for this PR, here comes my review :slightly_smiling_face: It'll be good to have this option now, but the best would be that this is handled in the integration with the Compiler so that mem copies can be overlapped with kernel launches.

pdroalves:
> Hi @pdroalves! Thank you for this PR, here comes my review slightly_smiling_face It'll be good to have this option now, but the best would be that this is handled in the integration with the Compiler so that mem copies can be overlapped with kernel launches.

Indeed. An option we have is to prepare this code for that by converting the allocation calls to their async versions. Without that, I don't think the Compiler will be able to overlap anything, anyway. Those calls force synchronization on the host side.

concrete-cuda/cuda/src/bootstrap_low_latency.cuh
pdroalves: That makes sense. Cooperative threads obligate us not to launch more blocks than we can run in parallel in the device. I don't know why, I assumed that a single block would dominate the SM, but this is not true.

I will run some tests to find a more tight condition.

concrete-cuda/cuda/src/bootstrap_low_latency.cuh
pdroalves: In case we can't run all samples at once, the buffer size only needs to support the amount of samples that will be processed in each iteration of the loop. Allocating more memory than that would be a waste of memory.

I think the problem here is the same as in the other comment, it assumes that we can only run a single sample per SM. This is not true.

agnesLeroy:
@pdroalves actually I was thinking of overlapping input data copies with kernel calls, but making the allocs async makes sense as well :slightly_smiling_face:

concrete-cuda/cuda/src/bootstrap_low_latency.cuh
pdroalves: I refactored that piece of code to check if `2 * remaining_samples * level_count < num_sm * active_blocks_per_sm`.The solution I used to compute the number of active blocks uses the rule you mentioned, but it is bounded to a particular hardware and doesn't adapt to possible modifications to the kernel.

An optimal solution should be calling cuOccupancyMaxActiveBlocksPerMultiprocessor on runtime to get that, but I couldn't figure out how to create the CUfunction object that function needs. I will need help to implement that.

concrete-cuda/cuda/src/bootstrap_low_latency.cuh
pdroalves: Fixed.

pdroalves:
@slab-ci gpu_test

pdroalves:
@slab-ci cpu_test

pdroalves:
@slab-ci gpu_test

pdroalves:
@slab-ci gpu_test

pdroalves:
@slab-ci cpu_test

concrete-cuda/cuda/src/bootstrap_low_latency.cuh
agnesLeroy: @pdroalves there's an example of use in the Cuda samples: https://github.com/NVIDIA/cuda-samples/blob/master/Samples/3_CUDA_Features/memMapIPCDrv/memMapIpc.cpp, maybe it can help?

PR #267, opened 2022-11-08
feat(cuda): add lwe / cleartext multiplication GPU acceleration

agnesLeroy:
### Checklist 


agnesLeroy:
@slab-ci cpu_test

agnesLeroy:
@slab-ci gpu_test

agnesLeroy:
@pdroalves I'll fix formatting after the tests have run sorry for this. :slightly_smiling_face:

concrete-core/src/backends/cuda/private/crypto/cleartext/list.rs
```diff
@@ -0,0 +1,110 @@
...
+    let input_slice = input.as_tensor().as_slice();
+    let count = CiphertextCount(input.count().0);
+    // In case there are less inputs than GPUs, we use just one GPU per input
+    let number_of_gpus = number_of_active_gpus(number_of_available_gpus, count);
+    let mut vecs = Vec::with_capacity(number_of_gpus.0);
+    let samples_on_gpu_0 = compute_number_of_samples_on_gpu(number_of_gpus, count, GpuIndex(0)).0;
+    for (gpu_index, chunk) in input_slice.chunks_exact(samples_on_gpu_0).enumerate() {
+        let stream = &streams[gpu_index];
+        let alloc_size =
+            compute_number_of_samples_on_gpu(number_of_gpus, count, GpuIndex(gpu_index)).0;
```
pdroalves, line 34: This part is a little confusing for me. Each iteration should process `chunk`, which has size `samples_on_gpu_0`. However, it also computes `alloc_size`, which is the number of samples that can be processed by some (maybe not-zero) GPU, allocates `alloc_size * sizeof(T)` bytes, and copy `chunk.len()` or `chunk.len() + samples_on_gpu_0` bytes. This can led to an undesired behavior.

I suggest this piece of code to be refactored so that each iteration receives `chunk` from a vector/slice and does all following computation considering only the size of that `chunk`.

pdroalves (changes requested):

pdroalves:
Hey @agnesLeroy !
The tests are passing and most of things look good to me. There is a single point that came to my eyes, which is the size used to allocate device memory not being 100% related to the vector being copied. I'm afraid that is also being done [for plaintexts](https://github.com/zama-ai/concrete-core/blob/main/concrete-core/src/backends/cuda/private/crypto/plaintext/list.rs).

Do you agree that's really a problem?

agnesLeroy:
> Hey @agnesLeroy ! The tests are passing and most of things look good to me. There is a single point that came to my eyes, which is the size used to allocate device memory not being 100% related to the vector being copied. I'm afraid that is also being done [for plaintexts](https://github.com/zama-ai/concrete-core/blob/main/concrete-core/src/backends/cuda/private/crypto/plaintext/list.rs).
> 
> Do you agree that's really a problem?

Why do you says it's not 100% related?

agnesLeroy:
@slab-ci gpu_test

pdroalves (approved):
Thanks for the modifications @agnesLeroy ! 
I think we have something much better now :-D

PR #268, opened 2022-11-08
chore(ci): enable avx512 builds in the FFI CI

IceTDrinker:
### Checklist 

### DIFF
--- a/.github/workflows/ffi_github_release.yml
+++ b/.github/workflows/ffi_github_release.yml
@@ -19,6 +19,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
       - name: Build libconcrete_core_ffi
         env:
           RUSTFLAGS: "-Ctarget-feature=+aes,+sse2,+avx,+avx2"
@@ -26,8 +31,8 @@ jobs:
           TMP_DIR="$(mktemp -d)"
 
           ./script/action_utils/concrete-core-ffi-release.sh \
-          --rust-toolchain "$(cargo xtask get_build_toolchain | tail -n 1)" \
-          --features-string '--features x86_64' \
+          --rust-toolchain +nightly \
+          --features-string '--features=x86_64,backend_fft_nightly_avx512' \
           --rust-build-dir "$(pwd)/target/release" \
           --output-release-dir "${TMP_DIR}" \
           --sha256-sum-out-file sha256_linux.txt \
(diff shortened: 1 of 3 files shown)


IceTDrinker:
We were discussing perf with @BourgerieQuentin and trying to enable avx512 on my laptop which does not have avx512 made no difference running the ffi, so enabling avx512 on x86 chips is safe thanks to some dynamic dispatch we do.

The upside is machines with avx512 will run concrete-core-ffi faster

.github/workflows/ffi_github_release.yml
```diff
@@ -49,15 +54,20 @@ jobs:
...
+          toolchain: nightly
+          override: true
       - name: Build libconcrete_core_ffi
         env:
           RUSTFLAGS: "-Ctarget-feature=+aes,+sse2"
         run: |
           TMP_DIR="$(mktemp -d)"
 
           ./script/action_utils/concrete-core-ffi-release.sh \
-          --rust-toolchain "$(cargo xtask get_build_toolchain | tail -n 1)" \
```
agnesLeroy, line 59: Why did we get the toolchain this way before?

concrete-core-ffi/CMakeLists.txt
```diff
@@ -1,8 +1,6 @@
 # concrete-core-ffi/CMakeLists.txt
 cmake_minimum_required(VERSION 3.16)
 project(concrete-core-ffi C)
-include_directories(${CONCRETE_BUILD_DIR})
-link_directories(${CONCRETE_BUILD_DIR})
```
agnesLeroy, line 5: Those lines were not used?

agnesLeroy (approved):
Thank you for this PR! I'm a bit confused because I thought we had activated this in the build some time ago but then reverted the change :thinking: but maybe it was something else.

concrete-core-ffi/CMakeLists.txt
IceTDrinker: Legacy from very old ffi, removing those lines does not cause issues

.github/workflows/ffi_github_release.yml
IceTDrinker: Basically to use avx512 we need the nightly toolchain, before we were using stable for x86 and nightly for aarch64, the C api is kind of an outlier now that all builds require a nightly toolchain

IceTDrinker:
actually after checking the code some more it probably does not emit avx512 instructions, I'll check the binaries and the asm in there

IceTDrinker:
so there is some avx512 code, it's not clear to me if it gets called however

PR #285, opened 2022-11-23
chore(ci): enable concrete-cuda benchmarks on slab

soonum:
### Description

Concrete-cuda benchmarks can be run either from a Pull Request or can be started using start_gpu_benchmark.yml.
On each push on main branch this workflow will be triggered. Results are parsed and sent to Slab CI server which will in turn make them available in our Grafana instance.

### Checklist 

### DIFF
--- a/ci/slab.toml
+++ b/ci/slab.toml
@@ -20,3 +20,8 @@ check_run_name = "AWS tests (Slab)"
 workflow = "aws_tests_gpu_slab_beta.yml"
 profile = "gpu"
 check_run_name = "AWS tests GPU (Slab)"
+
+[command.gpu_bench]
+workflow = "gpu_benchmark.yml"
+profile = "gpu"
+check_run_name = "AWS benchmark GPU (Slab)"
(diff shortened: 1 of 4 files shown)


ci/slab.toml
```diff
@@ -11,6 +11,13 @@ instance_type = "p3.2xlarge"
 subnet_id = "subnet-8123c9e7"
 security_group = "sg-0466d33ced960ba35"
 
+[profile.bench_cuda]
+region = "us-east-1"
+image_id = "ami-0ae662beb44082155"
+instance_type = "p3.16xlarge"
```
agnesLeroy, line 17: I think p3.2xlarge would be enough considering that we don't do multi GPU :slightly_smiling_face: -> p3.2xlarge instead

agnesLeroy (changes requested):

ci/slab.toml
soonum: I hope we won't have to much variation between runs then.

agnesLeroy (approved):

agnesLeroy (approved):

PR #295, opened 2022-11-30
fix(concrete-cuda): Do not expose cudaStream_t in public API

BourgerieQuentin:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=SnbZyQt)

The type cudaStream_t was used in device.h header file and never defined, that leads to error when the header file was include by the compiler runtime. I follow the same pattern than the other function that takes stream to avoid exposing cudaStream_t directly.

### DIFF
--- a/concrete-cuda/cuda/include/device.h
+++ b/concrete-cuda/cuda/include/device.h
@@ -7,7 +7,7 @@ int cuda_destroy_stream(void *v_stream, uint32_t gpu_index);
 
 void *cuda_malloc(uint64_t size, uint32_t gpu_index);
 
-void *cuda_malloc_async(uint64_t size, cudaStream_t stream, uint32_t gpu_index);
+void *cuda_malloc_async(uint64_t size, void* v_stream, uint32_t gpu_index);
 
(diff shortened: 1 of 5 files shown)


agnesLeroy:
@BourgerieQuentin have you tried to include 
```
#include <cuda_runtime.h>
```
in device.h instead? We made this choice to avoid casting the stream every time we call these functions.

BourgerieQuentin:
Yes I tried that but actually when it is included in the compiler runtime library the header file is not found. I will try to do that and fix the compiler cmake to include cuda include directories when we build with cuda backend. I let you known

agnesLeroy:
> Yes I tried that but actually when it is included in the compiler runtime library the header file is not found. I will try to do that and fix the compiler cmake to include cuda include directories when we build with cuda backend. I let you known

Have you defined the following environment variables in your build environment?
```
export CUDA_HOME="/usr/local/cuda"
export LD_LIBRARY_PATH=${CUDA_HOME}/lib64:$LD_LIBRARY_PATH
```
(the first path may vary depending on how cuda is installed)

PR #303, opened 2022-12-07
refactor(core): use from_le_bytes for gaussian RNG (see uniform RNG)

IceTDrinker:
- avoids small allocations, uses std::mem::size_of for size

### Checklist 


IceTDrinker:
Rewriting the code this way, I'm wondering to what extent the generation produces NaNs as we are interpreting bytes directly as floats, might be those values causing some test failures ? Unless there is something I'm missing which would avoid NaNs/convert to some special values

IceTDrinker:
Well it's possible but unlikely rust playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=231179040b1b961f753df120adfba64e

IceTDrinker:
actually the test > 0 and < 1 seems to be working properly with NaNs

IceTDrinker:
https://en.wikipedia.org/wiki/NaN#Comparison_with_NaN

IceTDrinker:
unless rust has optimizations for small vectors, the vec! call allocates on the heap.

Also we had a discussion a while back about using from_le_bytes to avoid our generators yielding different results on archs/platforms with different endianness

IceTDrinker:
I stumbled upon this code when looking for unsafe code in tfhe-rs, seemed ok to rewrite slightly to avoid memory trickery

agnesLeroy:
Alright makes sense, then it's all good for me!

PR #310, opened 2022-12-16
Fix/increase overflow

IceTDrinker:
### Checklist 

### DIFF
--- a/concrete-csprng/Cargo.toml
+++ b/concrete-csprng/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "concrete-csprng"
-version = "0.2.2"
+version = "0.2.3"
 edition = "2021"
 authors = ["D. Ligier", "J.B. Orfila", "A. Péré", "S. Tap", "Zama team"]
 license = "BSD-3-Clause-Clear"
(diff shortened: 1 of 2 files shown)


agnesLeroy (approved):

IceTDrinker:
so this is a genuine potential bug, turns out the cause of the bug I caught is overflows but in the forking methods :|

What do you think about switching some of those usize variables to plain u64 ?

agnesLeroy:
> so this is a genuine potential bug, turns out the cause of the bug I caught is overflows but in the forking methods :|
> 
> What do you think about switching some of those usize variables to plain u64 ?

I thought it wouldn't work on 32 bits platforms, hence why we had used usize everywhere :thinking:

IceTDrinker:
> > so this is a genuine potential bug, turns out the cause of the bug I caught is overflows but in the forking methods :|
> > What do you think about switching some of those usize variables to plain u64 ?
> 
> I thought it wouldn't work on 32 bits platforms, hence why we had used usize everywhere 🤔

Ah well, wasm is a 32 bits platform at the moment 😬

IceTDrinker:
I'll avoid using forking in wasm32 for now, it's a feature mainly to have reproducible multi/mono threaded keygens anyways

agnesLeroy:
Alright, then I'll merge this one, but we should take care of overflows in the forking at some point.

PR #323, opened 2023-02-07
refactor(cuda): Refactor the amortized PBS to support k > 1

pdroalves:
### Checklist 

### DIFF
--- a/concrete-core-fixture/src/fixture/lwe_ciphertext_vector_discarding_bootstrap_3.rs
+++ b/concrete-core-fixture/src/fixture/lwe_ciphertext_vector_discarding_bootstrap_3.rs
@@ -12,14 +12,14 @@ use concrete_core::prelude::{
     DecompositionBaseLog, DecompositionLevelCount, DispersionParameter, GlweCiphertextCount,
     GlweCiphertextVectorEntity, GlweDimension, LogStandardDev, LweBootstrapKeyEntity,
     LweCiphertextCount, LweCiphertextVectorDiscardingBootstrapEngine, LweCiphertextVectorEntity,
-    LweDimension, PolynomialSize, StandardDev, Variance,
+    LweDimension, PolynomialSize, Variance,
 };
 
 /// A fixture for the types implementing the `LweCiphertextVectorDiscardingBootstrapEngine` trait.
-pub struct LweCiphertextVectorDiscardingBootstrapFixture2;
+pub struct LweCiphertextVectorDiscardingBootstrapFixture3;
 
 #[derive(Debug)]
-pub struct LweCiphertextVectorDiscardingBootstrapParameters2 {
+pub struct LweCiphertextVectorDiscardingBootstrapParameters3 {
     pub noise: Variance,
     pub lwe_dimension: LweDimension,
     pub glwe_dimension: GlweDimension,
@@ -50,7 +50,7 @@ impl<
             InputCiphertextVector,
             OutputCiphertextVector,
         ),
-    > for LweCiphertextVectorDiscardingBootstrapFixture2
+    > for LweCiphertextVectorDiscardingBootstrapFixture3
 where
     Precision: IntegerPrecision,
     InputKeyDistribution: KeyDistributionMarker,
@@ -74,7 +74,7 @@ where
         + SynthesizesLweCiphertextVector<Precision, InputKeyDistribution, InputCiphertextVector>
         + SynthesizesLweCiphertextVector<Precision, OutputKeyDistribution, OutputCiphertextVector>,
 {
-    type Parameters = LweCiphertextVectorDiscardingBootstrapParameters2;
+    type Parameters = LweCiphertextVectorDiscardingBootstrapParameters3;
     type RepetitionPrototypes = (
         <Maker as PrototypesGlweCiphertextVector<Precision, OutputKeyDistribution>>::GlweCiphertextVectorProto,
         <Maker as PrototypesLweSecretKey<Precision, InputKeyDistribution>>::LweSecretKeyProto,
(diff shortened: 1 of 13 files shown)


pdroalves:
To add support for k>1 in the amortized PBS I had to refactor and fix things all over the crate, including some common methods to the low latency kernel.

pdroalves:
@slab-ci gpu_test

concrete-cuda/cuda/src/bootstrap_amortized.cuh
```diff
@@ -338,11 +295,11 @@ __host__ void host_bootstrap_amortized(
...
         <<<grid, thds, SM_FULL, *stream>>>(
             lwe_array_out, lut_vector, lut_vector_indexes, lwe_array_in,
-            bootstrapping_key, d_mem, input_lwe_dimension, polynomial_size,
-            base_log, level_count, lwe_idx, 0);
+            bootstrapping_key, d_mem, glwe_dimension, lwe_dimension,
+            polynomial_size, base_log, level_count, lwe_idx, 0);
   }
   checkCudaErrors(cudaGetLastError());
-
+  cudaDeviceSynchronize();
```
agnesLeroy, line 302: I think we should remove this synchronize from here: in the PR where I fixed the asynchronous behaviour I removed them all, and called them from the tests or the Rust integration instead.

concrete-cuda/cuda/src/crypto/gadget.cuh
```diff
@@ -11,46 +11,54 @@ private:
...
-      state[tid + params::degree / 2] >>= base_log;
-      T carry_re = ((res_re - 1ll) | state[tid]) & res_re;
-      T carry_im = ((res_im - 1ll) | state[tid + params::degree / 2]) & res_im;
-      carry_re >>= (base_log - 1);
-      carry_im >>= (base_log - 1);
-      state[tid] += carry_re;
-      state[tid + params::degree / 2] += carry_im;
-      res_re -= carry_re << base_log;
-      res_im -= carry_im << base_log;
+    for (int j = 0; j < num_poly; j++) {
```
agnesLeroy, line 38: Here I can see how the bootstrap code is simpler by doing the decomposition on every mask/body element in the same function, but we should probably change the function name to something like `decompose_and_compress_next_glwe` (name to be discussed) to make it clearer in the name that it's not only one value that's being dealt with. Also, this deserves an explanation in a comment for the function.

concrete-cuda/cuda/src/polynomial/functions.cuh
```diff
@@ -62,33 +62,38 @@ __device__ void add_polynomial_inplace_low_lat(T *source, T *dst, int p_id) {
...
-        if (tid < degree - j) {
-          accumulator[tid] = input[tid + j];
-        } else {
-          accumulator[tid] = -input[tid - degree + j];
-        }
-      } else {
-        uint32_t jj = j - degree;
-        if (tid < degree - jj) {
-          accumulator[tid] = -input[tid + jj];
+  for (int z = 0; z < (glwe_dimension + 1); z++) {
```
agnesLeroy, line 68: I have the same remark here: the function name should be adapted to reflect this change, and a comment should be added to make it clear that it handles all GLWE dimensions, not just one value.

concrete-cuda/cuda/src/polynomial/functions.cuh
```diff
@@ -98,28 +103,31 @@ __device__ void divide_by_monomial_negacyclic_inplace(T *accumulator, T *input,
...
-                                                   uint32_t j) {
-  int tid = threadIdx.x;
+__device__ void multiply_by_monomial_negacyclic_and_sub_polynomial(
+    T *acc, T *result_acc, uint32_t j, uint32_t glwe_dimension) {
   constexpr int degree = block_size * elems_per_thread;
-  for (int i = 0; i < elems_per_thread; i++) {
-    if (j < degree) {
-      if (tid < j) {
-        result_acc[tid] = -acc[tid - j + degree] - acc[tid];
+  for (int z = 0; z < (glwe_dimension + 1); z++) {
```
agnesLeroy, line 109: Same here.

concrete-cuda/cuda/src/polynomial/functions.cuh
```diff
@@ -129,19 +137,23 @@ multiply_by_monomial_negacyclic_and_sub_polynomial(T *acc, T *result_acc,
...
-    T b_acc = (x_acc & mask) >> (shift - 1);
-    T res_acc = x_acc >> shift;
-    res_acc += b_acc;
-    res_acc <<= shift;
-    rotated_acc[tid] = res_acc;
-    tid = tid + block_size;
+                                                  int level_count,
+                                                  uint32_t glwe_dimension) {
+  constexpr int degree = block_size * elems_per_thread;
+  for (int z = 0; z < (glwe_dimension + 1); z++) {
```
agnesLeroy, line 143: Same here again.

concrete-cuda/cuda/src/polynomial/functions.cuh
```diff
@@ -177,79 +189,88 @@ __device__ void add_to_torus(double2 *m_values, Torus *result) {
...
+    for (int i = 0; i < params::opt; i++) {
+      accumulator_slice[tid] = -accumulator_slice[tid];
+      tid = tid + params::degree / params::opt;
+    }
+    synchronize_threads_in_block();
 
-  // Reverse the accumulator
-  // accumulator.reverse_inplace();
+    // Reverse the accumulator
+    // accumulator.reverse_inplace();
```
agnesLeroy, line 219: We could remove those commented lines maybe?

concrete-cuda/cuda/src/polynomial/functions.cuh
```diff
@@ -177,79 +189,88 @@ __device__ void add_to_torus(double2 *m_values, Torus *result) {
...
+    for (int i = 0; i < params::opt; i++) {
+      accumulator_slice[tid] = result[i];
+      tid = tid + params::degree / params::opt;
+    }
+    synchronize_threads_in_block();
 
-  // Perform ACC * X
-  // accumulator.multiply_by_monomial_negacyclic_inplace(1);
+    // Perform ACC * X
+    // accumulator.multiply_by_monomial_negacyclic_inplace(1);
```
agnesLeroy, line 238: Maybe we could remove these commented lines as well?

concrete-cuda/cuda/src/polynomial/functions.cuh
```diff
@@ -177,79 +189,88 @@ __device__ void add_to_torus(double2 *m_values, Torus *result) {
...
+    for (int i = 0; i < params::opt; i++) {
+      accumulator_slice[tid] = result[i];
+      tid += params::degree / params::opt;
+    }
+    synchronize_threads_in_block();
 
-  // Copy to the mask of the LWE sample
-  // accumulator.copy_into(lwe_array_out);
+    // Copy to the mask of the LWE sample
+    // accumulator.copy_into(lwe_array_out);
```
agnesLeroy, line 266: Here again line 266 could be removed.

agnesLeroy (changes requested):
Hi @pdroalves! Thanks a lot for this PR, the code looks much more readable to me now! I have a few remarks you'll see, nothing major :slightly_smiling_face: You did an amazing job!

concrete-cuda/cuda/src/bootstrap_amortized.cuh
pdroalves: Ops, this is a debug artifact. I forgot to remove :sweat_smile:.

concrete-cuda/cuda/src/crypto/gadget.cuh
pdroalves: Something that is not clear to me is how generic this function is (as well as all functions in `functions.cuh`). As I see, this gadget class can be used to decompose any number of polynomials concatenated in an array. These can be a GLWE with any dimension, but also anything else. For instance, I have yet to start the refactoring of the low latency kernel, but I believe one path we could take there is to have different y-blocks handling each polynomial in the mask. In that case, we would need each `decompose_and_compress_next` to decompose a single polynomial and not the entire GLWE. So I'm not very supportive in using "glwe" there unless we have sure we will never use these functions to anything than on GLWEs. What do you think?

Otherwise, I think just adding a comment explaining the class would be enough.

concrete-cuda/cuda/src/polynomial/functions.cuh
pdroalves: These comments were quite useful to me to understand what is being done in each part of the function. I would keep them.

pdroalves:
Should we simply ignore the CPU test? I think this is one of those problematic statistical tests.

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_discarding_bootstrap_3.rs
```diff
@@ -0,0 +1,335 @@
...
+        criteria: &Self::Criteria,
+        outputs: &[Self::Outcome],
+    ) -> bool {
+        let (delta_max,) = criteria;
+        let (means, actual): (Vec<_>, Vec<_>) = outputs.iter().cloned().unzip();
+
+        for (x, y) in means.into_iter().zip(actual.into_iter()) {
+            for (expected, obtained) in x.into_iter().zip(y) {
+                if (<Precision::Raw as CastInto<i64>>::cast_into(expected)
+                    - <Precision::Raw as CastInto<i64>>::cast_into(obtained))
```
agnesLeroy, line 325: Is this not the same check as in the fixture 2? Why did you add the fixture 3?

concrete-cuda/cuda/src/crypto/gadget.cuh
agnesLeroy: Ok agreed, a comment should be enough for us to keep track of what we can do with the function.

concrete-cuda/cuda/src/polynomial/functions.cuh
agnesLeroy: I was confused about line 219 which is pseudo code, but we don't actually have a reverse_inplace() function in the code base. Maybe replacing the pseudo code with an explanation would be clearer?

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_discarding_bootstrap_3.rs
pdroalves: Fixture3 tests the PBS for different parameters, with k>1. I couldn't find a way to check which engine was being tested, so I couldn't avoid using those new parameters on the low latency kernel, which does not support k>1 yet.

agnesLeroy: Oh ok makes sense. It's a shame to add a new fixture then but we can remove it when low lat PBS supports k > 1 as well. No problem for me :slightly_smiling_face:

pdroalves:
I added a comment explaining how the `GadgetMatrix` works. Moreover, I also changed all the functions in `functions.cuh` to receive a number of concatenate polynomials instead of `glwe_dimension`, except for the `sample_*` methods since it makes sense to talk about GLWEs in that context.

pdroalves:
@slab-ci gpu_test

concrete-core-fixture/src/fixture/lwe_ciphertext_vector_discarding_bootstrap_3.rs
pdroalves: Totally! It is just temporary.

PR #337, opened 2023-02-27
fix(cuda): fix cleanup functions: it is necessary to sync the stream before freeing memory

agnesLeroy:
### Checklist 

### DIFF
--- a/concrete-cuda/cuda/src/bit_extraction.cu
+++ b/concrete-cuda/cuda/src/bit_extraction.cu
@@ -315,6 +315,7 @@ void cuda_extract_bits_64(void *v_stream, uint32_t gpu_index,
 void cleanup_cuda_extract_bits(void *v_stream, uint32_t gpu_index,
                                int8_t **bit_extract_buffer) {
   auto stream = static_cast<cudaStream_t *>(v_stream);
+  check_cuda_error(cudaStreamSynchronize(*stream));
   // Free memory
   cuda_drop_async(*bit_extract_buffer, stream, gpu_index);
 }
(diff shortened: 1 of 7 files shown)


concrete-cuda/cuda/src/bit_extraction.cu
```diff
@@ -315,6 +315,7 @@ void cuda_extract_bits_64(void *v_stream, uint32_t gpu_index,
 void cleanup_cuda_extract_bits(void *v_stream, uint32_t gpu_index,
                                int8_t **bit_extract_buffer) {
   auto stream = static_cast<cudaStream_t *>(v_stream);
+  check_cuda_error(cudaStreamSynchronize(*stream));
```
antoniupop, line 318: Is it actually necessary to synchronize the stream at this point (same for all cleanup functions below)? I thought that asynchronous deallocation was ordered in the stream so no need to wait for completion of stream tasks. 
I guess it's not a massive overhead as in optimised uses this would be only happening once after lots of uses, but it does make the call synchronous.

agnesLeroy: It's rather to make sure the stream is synchronized with the CPU. I didn't think we would need this synchronization here but it seems to be fixing my async setup in concrete-core :thinking: I'm running all the tests in my other PR (https://github.com/zama-ai/concrete-core/pull/335) to confirm it.

antoniupop: I tried synchronizing before dropping, but to no avail on my end. I'll try again just in case I missed something.

agnesLeroy: Hmm it doesn't seem to fix my tests though. They were running locally it's strange.

antoniupop: Yes I have tried again also using launch_blocking and still same for both configurations.

PR #338, opened 2023-02-27
chore(cuda): Reduces shared memory consumption in the amortized PBS and improves loop unrolling.

pdroalves:
### Description
By using the new `decompose_and_compress_next_polynomial()` we can shrink again the size of `accumulator_fft` to store a single polynomial. Moreover, this PR also replaces loops of the form

```
for (int i = 0; i < (glwe_dim+1) * params::opt; i++){
   ...
}
```

by
```
for (int i = 0; i < (glwe_dim+1); i++)
   for (int j = 0; j < params::opt; j++){
   ...
   }
```

so the compiler is capable of unrolling the inner loop.

### Checklist 

### DIFF
--- a/concrete-cuda/cuda/src/bootstrap_amortized.cuh
+++ b/concrete-cuda/cuda/src/bootstrap_amortized.cuh
@@ -124,11 +124,12 @@ __global__ void device_bootstrap_amortized(
     // The polynomial multiplications happens at the block level
     // and each thread handles two or more coefficients
     int pos = threadIdx.x;
-    for (int j = 0; j < (glwe_dimension + 1) * params::opt / 2; j++) {
-      res_fft[pos].x = 0;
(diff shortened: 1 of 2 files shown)


pdroalves:
@slab-ci gpu_test

agnesLeroy (approved):

pdroalves:
I updated the [spreadsheet](https://docs.google.com/spreadsheets/d/1Eh-WWsv_FJigsVvpS9a0FMlh_oNvB3fKVdt9zIqwwRM/edit#gid=142930041) with two new pages with data of this issue and the code in the current main branch. We have some speedups compared to the beginning of February and this issue in particular seems to improve that, which is good. However, we still have a slowdown when compared to January @agnesLeroy @bbarbakadze .

I will merge this PR since it improves performance, but we still have work to do.

PR #342, opened 2023-03-09
chore(cuda): Add checks to validate inputs passed to the scratch functions

pdroalves:
### Description

This commit moves all asserts from `host_*` methods to a single function and inserts it also in the scratch functions.

### Checklist 

### DIFF
--- a/concrete-cuda/cuda/src/bit_extraction.cu
+++ b/concrete-cuda/cuda/src/bit_extraction.cu
@@ -1,6 +1,42 @@
 #include "bit_extraction.cuh"
 
 /*
+ * Runs standard checks to validate the inputs
+ */
+void checks_fast_extract_bits(int nbits, int polynomial_size,
+                              int level_count_bsk, int number_of_samples) {
+
+  assert(("Error (GPU extract bits): polynomial_size should be one of "
+          "256, 512, 1024, 2048, 4096, 8192",
+          polynomial_size == 256 || polynomial_size == 512 ||
+              polynomial_size == 1024 || polynomial_size == 2048 ||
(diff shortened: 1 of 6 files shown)


pdroalves:
@slab_ci gpu-test

agnesLeroy (approved):
Looks really good, thanks a lot!

pdroalves:
@agnesLeroy bad news. I just noticed that actually this PR will be a little bigger than I expected. There are many parameters that are not passed to the scratch functions, as levels and base logs. I will have to modify the signature of several, which will break compatibility with many things. Do you think it is worthy?

agnesLeroy:
> @agnesLeroy bad news. I just noticed that actually this PR will be a little bigger than I expected. There are many parameters that are not passed to the scratch functions, as levels and base logs. I will have to modify the signature of several, which will break compatibility with many things. Do you think it is worthy?

Oh bad news indeed. No I don't think it's worthy: is it only the bit extract check that's concerned? If that's the case for that one we could have two different check functions, one for the scratch and one for the actual function. In the scratch we would only check what we can with the parameters we have. Wdyt?

pdroalves:
I think it is a better idea. We could have a "full" and a "fast" check version in those cases. I will adjust the code.

pdroalves:
@slab-ci gpu_test

agnesLeroy (commented):
Looks good to me, the cbs+vp check is missing though (in the wop_bootstrap.cuh file).

pdroalves:
Done. I've just pushed the missing checks.  I will launch the tests.

agnesLeroy (approved):

